PR #23, opened 2017-07-26
[WIP] Add faq back

nickjer:
We need to choose some questions from this list and provide adequate responses.

There may also be some recorded questions from the various webinars that may be great to include.


nickjer:
In case you don't want to look at the source:

screenshot from 2017-07-26 13-11-30

PR #28, opened 2017-07-27
OOD 1.0.0 Release

ericfranz:


ericfranz:
Ah so with GitFlow we create the release branch first, fix it to our liking, and then when we are ready for the world to see it, we merge that once into master. So I'll close this as well.

PR #29, opened 2017-07-27
OOD Release 1.0

ericfranz:


source/install-desktops.rst
```diff
@@ -0,0 +1,22 @@
+.. _install-desktops:
+
+Install Desktops
+================
+
+This installation guide will walk you through setting up an interactive Desktop
+App that your users will be able to use to launch an interactive Desktop
+session within a batch job on a compute node. The user should then be able to
+connect to a running session through their browser using the `noVNC`_ client.
```
ericfranz: I wonder if we should introduce the concept of "interactive HPC" here... Instead of "Install Desktops" should it be called "Install Interactive HPC Desktops"

source/install-desktops/customize-desktop-app.rst
```diff
@@ -0,0 +1,34 @@
...
+
+Customize Desktop App
+=====================
+
+In some cases the defaults used to submit a batch job may not work on your
+given cluster or you may just want to remove some form options presented to the
+user and fill in hard-coded values. All of these cases and more can easily be
+customized in the following files:
+
+:file:`bc_desktop/local/{cluster_desktop}.yml`
```
ericfranz: In the previous section we call this `cluster1.yml`, `cluster2.yml` and now its called `cluster_desktop.yml`. Will readers know that we are talking about the same file?

source/install-desktops/customize-desktop-app.rst
```diff
@@ -0,0 +1,34 @@
...
+  describes a given desktop app along with form attributes presented to the
+  user
+
+:file:`bc_desktop/local/submit/{custom_submit}.yml.erb`
+  describes how the batch job should be submitted to your cluster (needs to be
+  specified in a :file:`bc_desktop/local/{cluster_desktop}.yml` configuration
+  file)
```
ericfranz: Maybe better clarify this by saying something like: form config files under `bc_desktop/local/` must explicitly specify the custom submit config file under `bc_desktop/local/submit/` if a customized submit arguments should be used after a form submission

source/install-desktops/customize-desktop-app.rst
ericfranz: Also, for each yaml config file created under `bc_desktop/local/` a separate desktop app will be displayed to the user.

I don't know if this is clear from the text.

source/install-desktops/customize-desktop-app.rst
```diff
@@ -0,0 +1,34 @@
...
+
+.. note::
+
+   The ``.erb`` file extension will cause the YAML configuration file to be
```
ericfranz: we have to be careful here, because I think in the current implementation, if you omit the `.erb` extension these files are not found

source/install-desktops/install-desktop-app.rst
```diff
@@ -0,0 +1,103 @@
...
+      sudo scl enable rh-ruby22 -- rake install:bc_desktop
+
+#. Navigate to your OnDemand site, in particular the Dashboard App, and you
+   should see in the top dropdown menu "Interactive Apps" => "Cluster1
+   Desktop".
+
+   After choosing "Cluster1 Desktop" from the menu, you should be presented
+   with a form to submit a Desktop to the given cluster.
+
+   Submit a desktop and wait for it to run. If you see a Desktop start Running
```
ericfranz: How about "desktop session" for both instead of in some places using "Desktop" and others "desktop".

source/install-desktops/install-desktop-app.rst
```diff
@@ -0,0 +1,103 @@
...
+   .. code-block:: sh
+
+      cd ~/ood/src/apps/
+
+#. By default this app is not built and installed with all the other required
+   OOD apps. You will need to build this app explicitly with:
+
+   .. code-block:: sh
+
+      scl enable rh-ruby22 nodejs010 git19 -- rake build:bc_desktop
```
ericfranz: This throws me off. It seems that `rake build:bc_desktop` doing a git clone and nothing else. These are plugins. They don't need to be built. But maybe reviit later...

source/install-desktops/install-desktop-app.rst
```diff
@@ -0,0 +1,103 @@
+.. _install-desktops-install-desktop-app:
+
+Install Desktop App
+===================
+
+No we will go through installing the provided `bc_desktop`_ app as well as
```
ericfranz: 1. Fix typo: **Now we will install the...**
2. We don't mention why it is prefixed `bc_`. Maybe a separate section?

source/install-desktops/customize-desktop-app.rst
```diff
@@ -0,0 +1,34 @@
...
+Customize Desktop App
+=====================
+
+In some cases the defaults used to submit a batch job may not work on your
+given cluster or you may just want to remove some form options presented to the
+user and fill in hard-coded values. All of these cases and more can easily be
+customized in the following files:
+
+:file:`bc_desktop/local/{cluster_desktop}.yml`
+  describes a given desktop app along with form attributes presented to the
```
ericfranz: here we lowercase desktop app.... other places we upper case Desktop App....

source/install-desktops/customize-desktop-app/custom-job-submission.rst
```diff
@@ -0,0 +1,194 @@
...
+
+All `batch script options`_ are underneath the ``script`` configuration option.
+In particular since there is no option to modify number of nodes/cpus, we need
+to directly interact with the ``native`` command line arguments. This is
+specified as an array of :command:`qsub` arguments.
+
+If you would like to mimic how Torque handles ``bc_num_slots`` (number of
+**nodes**), then we will first need to change the form label of
+``bc_num_slots`` that the user sees in the form. This can be done by modifying
+our Desktop app local YAML configuration file:
```
ericfranz: Desktop App or desktop app?

source/install-desktops/customize-desktop-app/custom-job-submission.rst
```diff
@@ -0,0 +1,194 @@
...
+   title: "Cluster1 Desktop"
+   cluster: "cluster1"
+   submit: "submit/my_submit.yml.erb"
+
+   # Add the below option that allows us to modify attributes
+   attributes:
+     bc_num_slots:
+       label: "Number of nodes"
+
+Now when we go to the Desktop app form in our browser it will have the new
```
ericfranz: Desktop App or desktop app?

source/install-desktops/modify-cluster-configuration.rst
```diff
@@ -0,0 +1,53 @@
+.. _install-desktops-modify-cluster-configuration:
+
+Modify Cluster Configuration
+============================
+
+If you haven't already done so, you will need to first have a corresponding
+cluster configuration file for the cluster you intend to launch desktops on. It
+is recommended you follow the directions on :ref:`add-cluster-config`.
```
ericfranz: Mention that in order for this to work, job_adapter section in the cluster config needs to be filled out for the cluster.

source/install-desktops/install-desktop-app.rst
```diff
@@ -0,0 +1,103 @@
...
+   .. code-block:: sh
+
+      # Make the `local/` directory if it doesn't already exist
+      mkdir local
+
+      # Change our working directory to `local/`
+      cd local
+
+#. For each cluster we want to launch desktop on we will need a corresponding
+   YAML configuration file (e.g., ``cluster1.yml``, ``cluster2.yml``, ...) that
```
ericfranz: how about we call this a "form configuration file". Is the `.erb` optional or required?

source/install-desktops/enable-reverse-proxy.rst
```diff
@@ -0,0 +1,149 @@
...
+            basic:
+              # ...
+              # set_host: "host=$(hostname)"
+              set_host: "host=$(hostname -A | awk '{print $1}')"
+            vnc:
+              # ...
+              # set_host: "host=$(hostname)"
+              set_host: "host=$(hostname -A | awk '{print $1}')"
+
+#. We will update the Apache configuration file by adding ``Location``
```
ericfranz: above this, add a small header "Steps to enable reverse proxy:" because we are now in a new section after "Requirements:"

source/install-desktops/enable-reverse-proxy.rst
```diff
@@ -0,0 +1,149 @@
+.. _install-desktops-enable-reverse-proxy:
+
+Enable Reverse Proxy
+====================
+
+The reverse proxy will proxy a request to any specified host and port through
```
ericfranz: I think its safe to remove this sentence: 

>This is different than what is used for proxying to per-user NGINX processes through Unix domain sockets. 

Also the words "reverse proxy" could be linked externally to https://en.wikipedia.org/wiki/Reverse_proxy

source/install-desktops/enable-reverse-proxy.rst
```diff
@@ -0,0 +1,149 @@
...
+IP sockets. This is different than what is used for proxying to per-user NGINX
+processes through Unix domain sockets. This can be used to connect to Jupyter
+notebook servers, RStudio servers, VNC servers, and more... This is disabled by
+default as it can be a security risk if not properly setup using a good
+``host_regex``.
```
ericfranz: After this summary, a short explanation would be a good idea with examples of both node and rnode i.e. here is a really rough draft:

 when enabled making a get request to /node/host123/5321/foo/bar will proxy to a server running on host123 and port 5321, making a get request to that server with the URL /node/host123/5321/foo/bar. Making a get request ti /rnode/host123/5321/foo/bar will proxy to a server running on host123 and port 5321, making a get reqeust to that server with the URL /foo/bar. Thus it is named "rnode" because when accessing a server through /rnode/ any links the server produces in its response (to assets or other pages, etc.) would need to be relative to the server root. Any server responding to requests "node" would need to be configured to run at the suburi /node/host123/5321. noVNC is configured to use /rnode/, while Jupyter and RStudio servers would be configured to use /node/.

source/install-desktops/enable-reverse-proxy.rst
```diff
@@ -0,0 +1,149 @@
...
+     Host: n0691.ten.osc.edu:5432
+     Upgrade-Insecure-Requests: 1
+     ...
+
+   .. note::
+
+      As we don't have the simple server return anything to the browser, you
+      can ignore any errors or warnings you see in your browser.
+
+      You can get fancier using a Python ``SimpleHTTPServer``, but we leave
```
ericfranz: Is this line necessary or can we just delete it?

ericfranz:
Note the reason we should introduce the Interactive HPC terms is because in the Gateways 17 submission we use the same term. In fact, I should get a copy of that submission, as we might consider maintaining some consistency between the way we discuss interactive apps in that demo proposal and here.

source/install-desktops/customize-desktop-app.rst
nickjer: The `{...}` signifies it is a variable. Whereas previously I was using `cluster1.yml` as the actual file name.

Also I didn't want to confuse people thinking that the name of the yaml file under `bc_desktop/local` had to match the name of the file in cluster configs. As those names do not need to correspond to each other and I wanted to clear up any confusion that they might be the same file.

nickjer: > Also, for each yaml config file created under `bc_desktop/local/` a separate desktop app will be displayed to the user.
>
> I don't know if this is clear from the text.

I don't mention it in the installation text, feel free to write one up.

source/install-desktops/customize-desktop-app.rst
nickjer: Hah, guess I lowercased it because it is "a" desktop app and not "the" Desktop App. But I am fine upper casing it.

ericfranz:
Gateways 17 submission is https://osu.box.com/s/16vkc2fgc3j9h13arvn3svylf460bs0w. Relevant parts:

>A common scenario at HPC centers involves users accessing GUI applications through a remote VNC session. OnDemand supports this through a web interface to VNC applications running on compute nodes launched from within a batch job. These are known as interactive apps. 
>
>After logging in, the user is first presented with the Dashboard app. The Dashboard is a navigational aid to discover and launch other apps. Users can select the Interactive Apps view, where they can launch apps such as Ansys, Abaqus, COMSOL, and Paraview (see Figure 1) and, from there, create and connect to interactive sessions for each app.  
>
>For our tutorial, we will create a MATLAB interactive app and deploy it to the Dashboard.  

Interesting. We don't call out "Interactive HPC" and instead just say "Interactive App" throughout the paper. We also don't use the word "plugin" anywhere. Nevertheless, I think the concept of Interactive HPC might be an interesting thing to call out. And batch connect plugins as the technology behind it.

source/install-desktops/customize-desktop-app.rst
ericfranz: I think I'll do "Desktop app" everywhere.

source/install-desktops/customize-desktop-app.rst
nickjer: If you omit the `.erb` extension they will be found but they won't be processed with ERB.

source/install-desktops/install-desktop-app.rst
nickjer: bc = before christ

source/install-desktops/install-desktop-app.rst
nickjer: We currently implement it on the OOD Demo server with:

```sh
scl enable rh-ruby22 nodejs010 git19 -- rake build:bc_desktop OOD_SITE=osc
```

also the build does more than a `git clone`, it can change the repository as well as `git fetch` and `git checkout`.

source/install-desktops/install-desktop-app.rst
nickjer: I am fine with renaming it. The `.erb` is optional although we don't currently have a use-case for it at OSC in the form configuration.

source/install-desktops/install-desktop-app.rst
nickjer: Sure.

source/install-desktops/customize-desktop-app/custom-job-submission.rst
nickjer: Who knows?

source/install-desktops/customize-desktop-app/custom-job-submission.rst
nickjer: Your guess is as good as mine.

source/install-desktops/modify-cluster-configuration.rst
nickjer: I believe this is a requirement though for the core apps to work. So it may be too much detail for this, but your call.

source/install-desktops/enable-reverse-proxy.rst
nickjer: We can include that explanation, although there is one here in the documenation:

https://osc.github.io/ood-documentation/develop/infrastructure/ood-portal-generator/configuration.html#ood-portal-generator-configuration-configure-reverse-proxy

But we don't bring up `node` and `rnode` after that summary, so including it there may confuse them.

source/install-desktops/enable-reverse-proxy.rst
nickjer: Seems reasonable.

source/install-desktops/enable-reverse-proxy.rst
nickjer: No, because they will get an error message in the browser. So that can be a source of confusion if we don't explain what they are seeing.

nickjer: Unless you mean the sentence below that about `SimpleHTTPServer` then I am fine with removing that. GitHub doesn't do a good job marking the line of interest.

ericfranz:
```
ericfranz:ood-documentation efranz$ grep ovverr source -R
source/install-desktops/customize-desktop-app/modify-form-attributes.rst:   Caution must be taken if you decide to ovverride the ``form`` configuration
source/install-desktops/customize-desktop-app/modify-form-attributes.rst:ovverriding the ``label`` option for the ``bc_account`` attribute).
source/install-desktops/customize-desktop-app/modify-form-attributes.rst:ovverriding the ``help`` option for the ``bc_account`` attribute).
source/install-desktops/customize-desktop-app/modify-form-attributes.rst:equivalent option for this attribute (for the above example we are ovverriding
source/installation/software-requirements.rst:      You will then add the following to ovverride the default settings:
```

We need to use a spell checker. I could just manually fix these spelling mistakes, but then how would I know whatever spell checker we chose worked?

This looks like a possibility: https://bitbucket.org/dhellmann/sphinxcontrib-spelling and https://doughellmann.com/blog/2011/05/26/creating-a-spelling-checker-for-restructuredtext-documents/ and https://pypi.python.org/pypi/sphinxcontrib-spelling/

nickjer:
We'd have to constantly update the dictionary of the spell checker. I am fine with just fixing spelling mistakes when we come across them.

ericfranz:
Updating the dictionary of a spell checker is okay.

ericfranz:
I updated the "subsubsection" titles to follow format "ood-portal-generator (from v0.3.1 to v0.4.0)". But I'm keeping the versions in the title, because its appropriate: all of the information in that subsubsection is not for "ood-portal-generator" but for "ood-portal-generator (from v0.3.1 to v0.4.0)" - the changes made from those two versions.

It is easier to glance over the changes and see what was changed too. We can enhance in the future later.

PR #41, opened 2017-08-04
Render rst files with jinja2 renderer first

### LINKED ISSUE #26: Replace versions with variables
Use substitution in the reST documents for OOD infrastructure/app versions, for easier updating.

nickjer:
Fixes #26

Renders RST file with a Jinja2 renderer first.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -344,3 +344,37 @@ texinfo_documents = [
 # If true, figures, tables and code-blocks are automatically numbered if they
 # have a caption. At same time, the numref role is enabled.
 numfig = True
+
+# -- Custom configuration -------------------------------------------------
+
+from jinja2 import Environment, StrictUndefined
+
+# Use Jinja to parse RST files first
+def rstjinja(app, docname, source):
+    """
+    Render our pages as a jinja template for fancy templating goodness.
+    """
+    jinja_env = Environment(undefined=StrictUndefined)
+    src = source[0]
+    rendered = jinja_env.from_string(src).render(doc_context)
+    source[0] = rendered
+
+def setup(app):
+    app.connect('source-read', rstjinja)
+
+# Context used for jinja template
+doc_context = {
+    # OOD component versions
+    'ood_portal_generator_version':  '0.4.0',
+    'mod_ood_proxy_version':         '0.3.1',
+    'ood_auth_map_version':          '0.0.3',
+    'nginx_stage_version':           '0.2.1',
+    'ood_apps_installer_version':    '1.0.0',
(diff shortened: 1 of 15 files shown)


nickjer:
It should be noted that the built-in variable substitution provided by rst documentation doesn't work for code blocks.

So doing this...

```rst
.. code-block:: sh

   git checkout v|nginx_stage_version|
```

would fail because it doesn't process the variable substitution inside the code block.

ericfranz:
So jinja uses mustache and handlebars compatible delimiters, unless you are in a code block, and then you need `v{{` instead of `{{`?

source/updating/update-desktops.rst
```diff
@@ -19,7 +19,7 @@ command:
 
    grep '##' /var/www/ood/apps/sys/bc_desktop/CHANGELOG.md
    #=> ## [Unreleased]
-   #=> ## [0.1.1] - 2017-07-12
+   #=> ## [{{ desktops_version }}] - YYYY-MM-DD
```
ericfranz: `v{{`?

source/conf.py
```diff
@@ -344,3 +344,37 @@
...
+# -- Custom configuration -------------------------------------------------
+
+from jinja2 import Environment, StrictUndefined
+
+# Use Jinja to parse RST files first
+def rstjinja(app, docname, source):
+    """
+    Render our pages as a jinja template for fancy templating goodness.
+    """
+    jinja_env = Environment(undefined=StrictUndefined)
```
ericfranz: This should work. I didn't test it though.

```python
jinja_env = Environment(undefined=StrictUndefined, variable_start_string='{{{{', variable_end_string='}}}}')
```

I can't imagine how that would affect anything outside of this function.

nickjer: I still don't think we should modify the template delimiters as it would make working with it more confusing in the future. As we'd have to translate all documentation and tutorials on Jinja2 into the new formatting within our heads.

ericfranz: We are barely using Jinja2 here. Just for these version strings. I think its appropriate. Though presumably in the future we might want to set different strings for all the Jinja2 delimiters in case we add tutorials on building simple OOD apps using Python.

Though its one find/replace in files, so switching to this later will take no time at all. As long as we document the escape hatch. We could wait till we start adding documentation about app development that includes code with handlebars or mustache or jinja.

ericfranz: I'll ignore this for now and merge. But if we start documenting anything here that would require the documentation to be escaped for jinja delimiters I'm going to change all of them then to use three `{{{` prefixed and `}}}` suffixed.

PR #45, opened 2017-09-12
Added directions to install Jupyter

### LINKED ISSUE #44: Install Jupyter
Add short tutorial on installing the Jupyter app, as a main section under "Install Desktops", by making use of the app developer views.

nickjer:
Fixes #44

### DIFF
--- a/source/add-jupyter/customize-attributes.rst
+++ b/source/add-jupyter/customize-attributes.rst
@@ -1,0 +1,129 @@
+.. _add-jupyter-customize-attributes:
+
+Customize Attributes
+====================
+
+Now we will customize the app to work on a given cluster. Be sure that you
+walked through :ref:`add-jupyter-software-requirements` for the given cluster
+ahead of time.
+
+Modify Form File
+----------------
+
+The main responsibility of the ``form.yml`` file is for defining the attributes
+(their values or HTML form elements) used when generating the batch script.
+
+#. We will begin by adding a cluster for the Jupyter app to use. You do this by
+   editing the ``form.yml`` in your favorite editor as such:
+
+   .. code-block:: yaml
+
+      # form.yml
+      ---
+
+      ...
+
+      cluster: "my_cluster"
+
+      ...
+
+   where we replace ``my_cluster`` with a valid cluster that corresponds to a
+   cluster configuration file located under
+   :file:`/etc/ood/config/clusters.d/{cluster}.yml`.
+
+#. We will also edit the module(s) that is(are) required to be loaded within
+   our batch job to get a Jupyter Notebook Server running:
+
+   .. code-block:: yaml
+
+      # form.yml
+      ---
+
+      ...
+
+      attributes:
+
+        ...
+
+        modules: "python/3.5"
+
+      ...
+
+   where we replace ``python/3.5`` with a list of required modules for our
+   given cluster.
+
+#. **Optional** If we do NOT have the Conda extensions installed for the above
+   Python modules then we must disable it:
+
+   .. code-block:: yaml
+
+      # form.yml
+      ---
+
+      ...
+
+      attributes:
+
+        ...
+
+        conda_extensions: "0"
+
+      ...
+
+   They are enabled by default in this file.
+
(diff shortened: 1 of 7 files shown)


source/add-jupyter/software-requirements.rst
```diff
@@ -0,0 +1,35 @@
+.. _add-jupyter-software-requirements:
+
+Software Requirements
+=====================
+
+The Jupyter app requires the following software to be installed on the nodes
```
brianmcmichael: maybe say "on the **compute** nodes" here for clarification

source/add-jupyter/customize-attributes.rst
```diff
@@ -0,0 +1,129 @@
+.. _add-jupyter-customize-attributes:
+
+Customize Attributes
+====================
+
+Now we will customize the app to work on a given cluster. Be sure that you
+walked through :ref:`add-jupyter-software-requirements` for the given cluster
```
brianmcmichael: walked -> walk
or 
walked -> have walked

source/add-jupyter/customize-attributes.rst
```diff
@@ -0,0 +1,129 @@
...
+ahead of time.
+
+Modify Form File
+----------------
+
+The main responsibility of the ``form.yml`` file is for defining the attributes
```
brianmcmichael: Mention where this is. The root of the app folder?

source/add-jupyter/customize-attributes.rst
```diff
@@ -0,0 +1,129 @@
...
+
+      ...
+
+      cluster: "my_cluster"
+
+      ...
+
+   where we replace ``my_cluster`` with a valid cluster that corresponds to a
+   cluster configuration file located under
+   :file:`/etc/ood/config/clusters.d/{cluster}.yml`.
```
brianmcmichael: The brackets here don't render into sphinx html. Maybe they need to be escaped? As a result, it's unclear that you're directing people to use the name of the yml file. Maybe change the name of the file to `my_cluster.yml` to indicate that the aforementioned `my_cluster` variable will be replaced with this name.

source/add-jupyter/customize-attributes.rst
```diff
@@ -0,0 +1,129 @@
...
+   They are enabled by default in this file.
+
+.. note::
+
+   It is recommended you commit the changes you made to ``form.yml`` to `git`_:
+
+   .. code-block:: sh
+
+      # Stage the modified file
+      git add form.yml
```
brianmcmichael: See note above about mentioning where this file is located. If it's preexisting and needs to be edited, I think having a `git add` command here makes it feel like we're creating, rather than editing this file.

If it already exists, maybe just use `git commit form.yml -m 'updated form with cluster attributes'`

source/add-jupyter/modify-submit-parameters.rst
```diff
@@ -0,0 +1,169 @@
...
+.. note::
+
+   The ``bc_num_slots`` shown above located within the ERB syntax is the value
+   returned from web form for "Number of nodes". We check if it is blank and
+   return a valid number.
+
+Other
+`````
+
+For most of our adapters (aside from Torque) the ``native`` attribute is an
```
brianmcmichael: You should mention the purpose of this array further up, before the torque and slurm examples.

It's not clear right now what the array does in those cases, or where it gets plugged into the template. It's only when you get to here that it makes sense.

source/add-jupyter/customize-attributes.rst
nickjer: The brackets cause the word `cluster` to be emphasized indicating that it is a variable and should be replaced.

For more information, please see:

http://www.sphinx-doc.org/en/stable/markup/inline.html#role-file

brianmcmichael: I see it now. It's not very noticeable in practice though, and it's confusing because you're trying to tell people that "my_cluster" should be replaced with "{cluster}.yml". I think the terms should be the same at least.

cluster

PR #46, opened 2017-09-19
App development

ericfranz:


README.md
```diff
@@ -18,6 +18,25 @@ built using the following command from the root of this repo:
...
+# then in the documentation root directory:
+WORKDIR=/doc PIPENV_VENV_IN_PROJECT=1 pipenv install
+```
+
+When building the html to test, run this command:
+
+```bash
+WORKDIR=/doc PIPENV_VENV_IN_PROJECT=1 pipenv run make html
+```
+
```
brianmcmichael: I was able to get this built on windows without docker using only `pip`, I can type up some simpler instructions to get this built without the additional dependency. Also, `homebrew` is macOS specific, so it won't work on linux or windows installations. This may need a header that mentions that these instructions are for macOS users only.

source/app-development/tutorial-ps-to-quota.rst
```diff
@@ -0,0 +1,451 @@
...
+     - contains setup code common between all tests
+
+Clone and setup
+---------------
+
+#. Login to Open OnDemand, click "Develop" dropdown menu and click the "My Sandbox Apps (Development)" option.
+#. Click "New Product" and "Clone Existing App".
+#. Fill out the form:
+
+   #. Directory name: quota
```
brianmcmichael: Use 'quota' so that it's clear that it's something the user will type.

source/app-development/tutorial-ps-to-quota.rst
```diff
@@ -0,0 +1,451 @@
...
+Update test/test_command.rb
+...........................
+
+Run the command to get example data. Copy and paste the output into the test, and
+update the assertions to expect an array of "quotas" instead of "processes"
+with appropriate attributes.
+
+Diff:
+
+.. code:: diff
```
brianmcmichael: Sphinx doesn't seem to handle this very well. I wonder if there's a different code for github markdown style diffs?

diff

source/app-development/tutorial-ps-to-quota.rst
```diff
@@ -0,0 +1,451 @@
...
+   With SCL, running git commands using git19 looks like:
+
+   ``scl enable git19 -- git commit -m "initial commit"``
+
+   You can avoid this by loading the SCL packages in your .bashrc or .bash_profile file.
+   For example, in my .bash_profile I have:
+
+   .. code:: sh
+
+      if [[ ${HOSTNAME%%.*} == webtest04*  ]]
```
brianmcmichael: Maybe mention above that `webtest04` should be replaced with the hostname of the system they are using.

source/app-development/tutorial-ps-to-quota.rst
```diff
@@ -0,0 +1,451 @@
...
+
+#. Updating the manifest.yml to specify the category and optionally subcategory, which indicates where in the dashboard menu the app appears.
+#. Having an administrator checkout a copy of the production version to a directory under /var/www/ood/apps/sys
+
+
+Steps:
+
+#. Add category to manifest so app appears in Files menu:
+
+    .. code:: diff
```
brianmcmichael: Adother diff, it looks like instead of using `code::`, sphinx has a `:diff:` option. All of the diffs should be replaced with this if it works.

See  http://www.sphinx-doc.org/en/stable/markup/code.html

source/app-development/tutorial-ps-to-quota.rst
```diff
@@ -0,0 +1,451 @@
...
+.. code:: diff
+
+    -name: Passenger App Processes
+    -description: Display your running Passenger app proceseses in a table
+    +name: Quota
+    +description: Display quotas
+    +icon: fa://hdd-o
+
+* The icon follows format of ``fa://{FONTAWESOMENAME}`` where you replace ``{FONTAWESOMENAME}`` with an icon from http://fontawesome.io/icons/.
+  In this case we are using ``fa-hdd-o`` which we write in the manifest as ``fa://hdd-o``: http://fontawesome.io/icon/hdd-o/
```
brianmcmichael: This looks a little funny in the build. Were you trying to hyperlink `fa-hdd-o` to the fontawesome site? The code looks like you were going for a rst hyperlink and the link seems kind of appendageal at the end.

brianmcmichael: fa

source/app-development/tutorial-ps-to-quota.rst
```diff
@@ -0,0 +1,451 @@
...
+#. Reload the dashboard.
+
+.. figure:: /images/app-dev-tutorial-ps-to-quota-published.png
+   :align: center
+
+   Every user can now launch the Quota from the Files menu.
+
+.. warning::
+
+   Accessing this new app for the first time will cause your NGINX server to restart, killing all websocket connections, which means resetting your active Shell sessions.
```
brianmcmichael: "which means any active web-based Shell sessions will reset!"

Personal preference make it active and add mention that this only applies to web-based shells.

source/app-development/tutorial-ps-to-quota.rst
nickjer: Try using...

```rst
.. code-block:: diff

   stuff...
```

instead of `code`.

ericfranz:
I applied most of the recommended changes. A few that I didn't I marked to do this weekend or next week but shouldn't hold up release (like updating the README).

PR #49, opened 2017-10-06
Authentication documentation

ericfranz:
**TODO: Before pulling in, revert eeb181e**

Tutorial can be previewed at https://osc.github.io/ood-documentation/release-1.1-auth/authentication/tutorial-oidc-keycloak-rhel7.html

Current problems with this documentation:

1. Need to determine which files (and/or directories) to protect access to in Keycloak so I can do 0640 or 0600 with keycloak:keycloak ownership. I can't do that for the entire directory because then I can cd into the directory to do work, and every single action requires sudo -u keycloak prefix unless I can easily sudo su -u keycloak, which I cannot.
2. The LDAP configuration section is messy with a bunch of OSC specific stuff, and it does make me nervous. After configuring LDAP, it would be great if there were directions I could give on testing and verifying that the LDAP configuration is correct (besides just the "test ldap connection" button)
3. In the Web UI setup, is Web Origins step required? In our directions we insert it, but I see in idp-dev for AweSim this is unset.
4. **Change the --password keycloakpass to be something that the user would just not copy/paste in!**

The more we can automate for this, the better. And for the non-automate-able steps, if we can provide tighter feedback loops, that would make this setup much easier.


treydock:
A few issues:

Install section should mention doing something like `cd /opt/keycloak-3.1.0.Final` since many of the commands being executed are relative paths.

Just preference but I prefer the following over doing vim + copy + paste.  This is more useful for things like the keycloak.service file where modifications are not likely.  Having vim + contents of things like Apache config make more sense since those will have to be modified by a site.

```
cat > /path/to/file <<EOF
CONTENTS
EOF
```

Note that `<<EOF` would need to be `<<'EOF'` if the content contains dollar signs that should not be expanded as variables.

Did you verify the install of mod_auth_openidc with those steps?  In my RPM spec file the quotes for `APXS2_OPTS` are single quotes, not double, which does matter.  If the steps you documented did work, after you uninstalled httpd24-mod_auth_openidc, then go with your steps.

Your note about RPM for mod_auth_openidc should say "Apache 2.4" instead of "httpd24" since the RPM is technically built against Apache 2.4 (OS) and not httpd24 (SCL).

For the theme, I think a link to Keycloak docs would be enough and maybe a link to the OSC theme as an example.

PR #75, opened 2018-01-02
Add queue drop-down tutorial for interactive apps

nickjer:
You can view built documentation at:

https://osc.github.io/ood-documentation-test/add-queue-dropdown/

### DIFF
--- a/source/app-development/tutorials-interactive-apps.rst
+++ b/source/app-development/tutorials-interactive-apps.rst
@@ -1,0 +1,13 @@
+.. _app-development-tutorials-interactive-apps:
+
+Tutorials: Interactive Apps
+===========================
+
+A curated list of tutorials for developing interactive apps.
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Tutorials
+
+   tutorials-interactive-apps/add-jupyter
+   tutorials-interactive-apps/add-custom-queue
--- a/source/app-development/tutorials-interactive-apps/add-custom-queue/global-static-list.rst
+++ b/source/app-development/tutorials-interactive-apps/add-custom-queue/global-static-list.rst
@@ -1,0 +1,251 @@
+.. _app-development-tutorials-interactive-apps-add-custom-queue-global-static-list:
+
+Use a Global Static List
+========================
+
+.. warning::
+
+   Requires permissions to modify the global cluster configurations located
+   under::
+
+     /etc/ood/config/clusters.d/
+
+You can take advantage of the Cluster Configuration files for defining a
+**common** static queue/partition list that can be used across all properly
+configured Interactive Apps. This is accomplished by:
+
+#. :ref:`app-development-tutorials-interactive-apps-add-custom-queue-global-static-list-modify-cluster-configuration`
+#. :ref:`app-development-tutorials-interactive-apps-add-custom-queue-global-static-list-add-custom-attribute`
+#. :ref:`app-development-tutorials-interactive-apps-add-custom-queue-global-static-list-handle-custom-attribute`
+
+.. _app-development-tutorials-interactive-apps-add-custom-queue-global-static-list-modify-cluster-configuration:
+
+Modify Cluster Configuration
+----------------------------
+
+.. danger::
+
+   This assumes you have walked through :ref:`add-cluster-config` for the
+   cluster you intend on submitting the Interactive App to.
+
+We will add a ``custom:`` field to our cluster configuration file if it doesn't
+already exist and then introduce a custom field (named anything your heart
+desires) with a list of available queues/partitions:
+
+.. code-block:: yaml
+   :emphasize-lines: 10-
+
+   # /etc/ood/config/clusters.d/my_cluster.yml
+   ---
+   v2:
+     metadata:
+       title: "My Cluster"
+     login:
+       host: "my_cluster.my_center.edu"
+     job:
+       # ... resource manager specific options here ...
+     custom:
+       queues: # this can be named anything your heart desires
+         - "queue1"
+         - "queue2"
+
+.. warning::
+
+   Be sure to click "Restart Web Server" on the Dashboard everytime you edit
+   the Cluster Configuration file for changes to take effect within the app.
+
+.. _app-development-tutorials-interactive-apps-add-custom-queue-global-static-list-add-custom-attribute:
+
+Add Custom Attribute to Form
+----------------------------
+
+We want to **replace** the ``bc_queue`` form attribute with a custom HTML
+``<select>`` element (a drop-down list of options) built up of queue/partition
+options read from the corresponding cluster configuration file.
+
+.. note::
+
+   You can read more about customizing attributes in the ``form.yml`` file for
+   Interactive Apps under the
+   :ref:`interactive-development-form-customizing-attributes` section.
+
+#. First we rename the ``form.yml`` for the Jupyter Interactive App to
+   ``form.yml.erb``:
+
+   .. code-block:: shell
+
+      mv ~/ondemand/dev/jupyter/form.yml ~/ondemand/dev/jupyter/form.yml.erb
+
+   This will cause the YAML file to be processed using the `eRuby (embedded
+   Ruby)`_ templating system, which allows us to embed Ruby_ code into the YAML
+   configuration file for flow control, variable substitution, and more.
+
+#. Next we start with the default ``form.yml`` for the Jupyter Interactive App:
+
+   .. code-block:: yaml
+
+      # ~/ondemand/dev/jupyter/form.yml.erb
+      ---
+      cluster: "my_cluster"
+      attributes:
+        modules: "python"
+        conda_extensions: "1"
+        extra_jupyter_args: ""
+      form:
+        - modules
+        - conda_extensions
+        - extra_jupyter_args
+        - bc_account
+        - bc_queue
+        - bc_num_hours
+        - bc_num_slots
+        - bc_email_on_started
+
+#. We remove the following line from this file:
+
+   .. code-block:: yaml
+      :emphasize-lines: 2
+
+      - bc_account
+      - bc_queue
+      - bc_num_hours
+
+   Now when we refresh the web page for our sandbox Jupyter App we won't see
+   the "Queue" form element anymore.
+
+#. We now add in code that reads in the list of available queues/partitions
+   from the cluster configuration file and generates a custom drop-down
+   attribute with this list of queues/partitions:
+
+   .. code-block:: yaml
+      :emphasize-lines: 2-4,11-18,24
+
+      # ~/ondemand/dev/jupyter/form.yml.erb
+      <%-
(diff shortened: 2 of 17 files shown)


source/app-development/tutorials-interactive-apps.rst
```diff
@@ -0,0 +1,13 @@
...
+===========================
+
+A curated list of tutorials for developing interactive apps.
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Tutorials
+
+   tutorials-interactive-apps/add-jupyter
+   tutorials-interactive-apps/add-custom-queue
```
ericfranz: I notice when I navigate between the subsections of add-custom-queue the navigation on the left disappears because on a new page you are scrolled to the top and you have to scroll down to see. The readthedocs theme I think supports "subprojects" https://docs.readthedocs.io/en/latest/faq.html#how-do-i-host-multiple-projects-on-one-cname (or that might just be the readthedocs site itself...). My guess is this would result in the left navigation being devoted entirely to that subproject, which could be "App Development" or even "Interactive App Development". This is outside the scope of this PR though, maybe worth opening another issue.

source/app-development/tutorials-interactive-apps/add-custom-queue/global-static-list.rst
```diff
@@ -0,0 +1,251 @@
...
+        - bc_account
+        - bc_queue
+        - bc_num_hours
+        - bc_num_slots
+        - bc_email_on_started
+
+#. We remove the following line from this file:
+
+   .. code-block:: yaml
+      :emphasize-lines: 2
```
ericfranz: What about using the diff formatting instead of just yaml and emphasize? In https://osc.github.io/ood-documentation-test/add-queue-dropdown/app-development/tutorials-passenger-apps/ps-to-quota.html#edit-to-run-and-parse-quota I split it into two blocks of code, showing the code diff, and then showing the final block with the code formatting. I probably could update https://osc.github.io/ood-documentation-test/add-queue-dropdown/app-development/tutorials-passenger-apps/ps-to-quota.html#edit-to-run-and-parse-quota to use emphasize-lines for the lines that were changed/modified like you do here though.

source/app-development/tutorials-interactive-apps/add-custom-queue/global-static-list.rst
```diff
@@ -0,0 +1,251 @@
...
+        modules: "python"
+        conda_extensions: "1"
+        extra_jupyter_args: ""
+        custom_queue:
+          label: Queue
+          help: Please select a queue from the drop-down.
+          widget: select
+          options:
+          <%- queues.each do |q| -%>
+            - [ "<%= q %>", "<%= q %>" ]
```
ericfranz: Good time to mention https://github.com/OSC/ood_core/issues/65. I didn't add this to ingest yet, but thought it was relevant considering this example.

source/app-development/tutorials-interactive-apps/add-custom-queue/global-static-list.rst
```diff
@@ -0,0 +1,251 @@
...
+      correspond to a cluster with a cluster configuration file. Also it should
+      match the cluster defined on the line:
+
+      .. code-block:: yaml
+
+         cluster: "my_cluster"
+
+      in the ``form.yml.erb`` file above.
+
+   Later in the YAML file we have another *expression tag* that loops through
```
ericfranz: When you say "execution tag" in parens probably should add (i.e. <%- and -%>). Also, I'm confused, does "execution tag" mean `<%-` and "expression tag" mean `<%=`.? In the examples you only use `<%-`.

Also, in documentation I see `<%` but do not see  `<%-`. Why do you use  `<%-`? I know that `%>` and `-%>` differs in that  `-%>` drops trailing whitespace. For example, neither https://ruby-doc.org/stdlib-2.5.0/libdoc/erb/rdoc/ERB.html#method-c-new nor http://api.rubyonrails.org/classes/ActionView/Base.html mention `<%-` but they do mention `-%>` i.e.

> You trigger ERB by using embeddings such as `<% %>`, `<% -%>`, and `<%= %>`.

brianmcmichael:
I don't see this at https://osc.github.io/ood-documentation-test/add-queue-dropdown/

source/app-development/tutorials-interactive-apps.rst
nickjer: Is this due to a smaller window? I am not seeing this on my window:

image

I can explore sub-projects, but like you mentioned it might be better to open a new issue for that (I can open that issue if you want, or you can). Not sure how time-consuming that exploration may be, and I know you want this out before the webinar.

source/app-development/tutorials-interactive-apps/add-custom-queue/global-static-list.rst
nickjer: I originally had diff, but it was a little confusing when I would glance at it because each of the lines had dashes already, but that one line had two dashes.

So it would look like:

> We remove the following line from this file:
> 
> ```diff
>   - bc_account
> - - bc_queue
>   - bc_num_hours
> ```

I am not married to either solution, I just went with emphasize because it was identical to the diff but didn't have that extra dash that "might" confuse someone and the difference in color (although I wish I could alter the emphasize color).

If you want me to go back to using diff I can.

source/app-development/tutorials-interactive-apps/add-custom-queue/global-static-list.rst
nickjer: Good points, I should fix those.

> Why do you use `<%-`

This strips leading whitespace:

```ruby
[1] pry(main)> require 'erb'
=> true
[2] pry(main)> ERB.new(" <% true -%>\nword", nil, "-").result
=> " word"
```

source/app-development/tutorials-interactive-apps/add-custom-queue/local-static-list.rst
```diff
@@ -0,0 +1,168 @@
...
+#. We remove the following line from this file:
+
+   .. code-block:: yaml
+      :emphasize-lines: 2
+
+      - bc_account
+      - bc_queue
+      - bc_num_hours
+
+   Now when we refresh the web page for our sandbox Jupyter App we won't see
```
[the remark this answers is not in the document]
nickjer: I can remove the mention of `sandbox` unless you have a better word, I was worried they might think of the system installed Jupyter App.

But I do mention `sandbox` elsewhere in the documentation, for example in the Jupyter App Tutorial:

https://osc.github.io/ood-documentation/master/app-development/add-jupyter/copy-app.html

There are also other places:

https://osc.github.io/ood-documentation/master/search.html?q=sandbox&check_keywords=yes&area=default

source/app-development/tutorials-interactive-apps/add-custom-queue/local-static-list.rst
```diff
@@ -0,0 +1,168 @@
...
+        modules: "python"
+        conda_extensions: "1"
+        extra_jupyter_args: ""
+        custom_queue:
+          label: Queue
+          help: Please select a queue from the drop-down.
+          widget: select
+          options:
+            - [ "queue1", "queue1" ]
+            - [ "queue2", "queue2" ]
```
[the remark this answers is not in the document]
nickjer: Unfortunately `serial` and `parallel` aren't common queues outside of OSC. I could use `debug` and `regular`.

source/app-development/tutorials-interactive-apps.rst
ericfranz: Yeah smaller window. Make be browser height small enough so "Use a Local Static List" in the left nav doesn't appear without you scrolling.

That said, absolutely, open a new issue, don't do it here, and lets not even think about touching that issue till Feb.

source/app-development/tutorials-interactive-apps/add-custom-queue/local-static-list.rst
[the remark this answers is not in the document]
ericfranz: Lets keep it.

source/app-development/tutorials-interactive-apps/add-custom-queue/local-static-list.rst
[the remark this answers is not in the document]
ericfranz: Maybe just capitalize queue1 and queue2 for the label i.e.:

```yaml
options:
  - [ "Queue1", "queue1" ]
  - [ "Queue2", "queue2" ]
```

At a glance then you can see one is the label, one is the value.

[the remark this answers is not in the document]
ericfranz: Also if my suggestion produces other problems than ignore my suggestion.

source/app-development/tutorials-interactive-apps/add-custom-queue/global-static-list.rst
ericfranz: I think the diff only works well if you call out that you are showing a diff, and works much better if you show removal and addition i.e. see the top of this as an example:

https://osc.github.io/ood-documentation-test/add-queue-dropdown-alt/app-development/tutorials-interactive-apps/add-custom-queue/local-static-list.html

But I do see value in showing removing the field and reloading as a separate step, to see the field being removed. Regardless, instead of saying

> We remove the following line from this file:

I would be more explicit and say:

> We remove the bc_queue field from this file:

When I was reading through the plain text, I saw:

>We remove the following line from this file:
>
>```
>- bc_account
>- bc_queue
>- bc_num_hours
>```

And I asked "why are we removing bc_account, maybe that was a mistake".

ericfranz:
@brianmcmichael In the navigation, click App Development, Tutorials: Interactive Apps, Add Custom Qeueus/Partitions

https://osc.github.io/ood-documentation-test/add-queue-dropdown/app-development/tutorials-interactive-apps/add-custom-queue.html

Most of the comments I added are minor enhancements or out of scope for this PR, so whatever you and @nickjer decide is good will be good enough for me.

source/app-development/tutorials-interactive-apps/add-custom-queue/global-static-list.rst
ericfranz: Also, I trust whatever you decide on this and don't need to review it again.

source/app-development/tutorials-interactive-apps/add-custom-queue/local-dynamic-list.rst
```diff
@@ -0,0 +1,240 @@
...
+   (e.g., Slurm, Torque, ...) agnostic.
+
+   .. note::
+
+      For the queue/partition we do not need to use the ``native:`` field which
+      **must be** customized for the specific resource manager you are
+      leveraging.
+
+      You can find a list of generic fields that are resource manager agnostic
+      under the `OodCore::Job::Script <script>`_ documentation.
```
ericfranz: This was a broken link and took me to https://osc.github.io/ood-documentation-test/add-queue-dropdown/app-development/tutorials-interactive-apps/add-custom-queue/script

PR #84, opened 2018-05-23
Add VMD to list of interactive apps [DO NOT MERGE]

laser505t:
What did I do wrong to make it think that I changed every file?


nickjer:
Haha, this is funny. So it seems your editor is adding the carriage return `\r` character to all the files. Windows treats newlines as `\r\n` and OSX/Linux treat it as just `\n`. Fortunately there is a setting in Git that can remove those carriage returns before it pushes it up to the repo:

https://help.github.com/articles/dealing-with-line-endings/#platform-windows

I'd recommend configuring your local git installation and then cloning this repo down again. Copy over the changes you made on a new branch and make a new PR. Also, you should name the branch `add-vmd` instead of `add-vms`.

nickjer:
I am not entirely sure how git is handling the line endings when it first cloned the repo to say if a `git revert` would work, since this was something internal to the git installation and not something Tyler did specifically, I thought it would be safest just to re-clone the repo after changing the configuration.

But yeah, it would have been better just to run a `git revert` and then squashed the PR during the merge in the end if it was a simple mistake.

ericfranz:
Closing as duplicate of https://github.com/OSC/ood-documentation/pull/86

PR #87, opened 2018-05-25
Change to console format

### LINKED ISSUE #81: Using 'console' format for installation directions
1. Use console for everything
2. Take advantage of the fact that the lexer puts `<span class="gp">$</span>` around the `$` and apply custom style `user-select: none` and `padding-right: 5px` to the `.highlight-console span.gp` class 
3. Remove the proceeding whitespace to the command, so it is typed in as `$tree /etc/ood/config`

The result is this:

select

#### Original full issue description

Using "shell" format without preceding `$` for commands users have to enter to do things makes it easier to copy a line: just click 3 times to highlight the line. This does copy the newline so if your shell isn't configured to not auto-execute the command when a newline is placed in (i.e. bash vs zsh) it could be bad if you want to paste, then inspect, then execute. But otherwise it is very convenient to follow through a set of directions.

One approach is to use console format with preceeding `$` for read only with lots of output. For example, in https://osc.github.io/ood-documentation/master/release-notes/v1.3-release-notes.html#upgrading-from-v1-2 we use shell formatting for all the commands you will copy and execute, where as we use the console formatting for `$ tree /etc/ood/config` because it conveniently formats the code below it.

Two problems  with this approach:

1. One problem is that this is inconsistent and if you are copying and pasting verbatim, as in testing an install in a VM, you might accidentally enter this full string `$ tree /etc/ood/config` instead of `tree /etc/ood/config`.
2. In the installation instructions there are many places where users enter commands AND we want to display the output.

What do other sites do? They are all over the place.

1. https://docs.pipenv.org/ just puts `$ ` in the example code
2. https://theforeman.org/manuals/1.1/index.html#3.2.1Installation displays commands shell like, without displaying output
3. Digital Ocean **displays commands and output in separate boxes** which is an option https://www.digitalocean.com/community/tutorials/initial-server-setup-with-ubuntu-16-04
4. Heroku uses a cool approach where the styling looks like "console" with the $ but the doubleclick actually highlights only the first line. This is because they are using https://developer.mozilla.org/en-US/docs/Web/CSS/user-select on the `$ `. Example: https://devcenter.heroku.com/articles/exec

The ideal might be to:

1. Use console for everything
2. Take advantage of the fact that the lexer puts `<span class="gp">$</span>` around the `$` and apply custom style `user-select: none` and `padding-right: 5px` to the `.highlight-console span.gp` class 
3. Remove the proceeding whitespace to the command, so it is typed in as `$tree /etc/ood/config`

The result is this:

select

laser505t:
Fixes #81

### DIFF
--- a/source/_static/css/custom.css
+++ b/source/_static/css/custom.css
@@ -23,5 +23,4 @@ span.gp {
   -moz-user-select: none;
   -ms-user-select: none;
   user-select: none;
-  padding-right: 5px;
 }
--- a/source/_static/css/custom.css
+++ b/source/_static/css/custom.css
@@ -17,3 +17,11 @@
 .wy-nav-top {
   background: #C8102E;
 }
+
+span.gp {
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  user-select: none;
+  padding-right: 5px;
+}
--- a/source/app-development/interactive/template.rst
+++ b/source/app-development/interactive/template.rst
@@ -193,7 +193,7 @@ An example before hook ``template/before.sh`` that defines a port and password
 file that can later be used in the script that launches the web server may look
 like:
 
-.. code-block:: shell
+.. code-block:: sh
 
    # Find available port to run web server on
    export port=$(find_port)
--- a/source/app-development/tutorials-passenger-apps/ps-to-quota.rst
+++ b/source/app-development/tutorials-passenger-apps/ps-to-quota.rst
@@ -134,15 +134,15 @@ Edit to run and parse quota
 
 The app runs and parses this command:
 
-.. code-block:: sh
+.. code-block:: console
 
-   ps aux | grep '[A]pp'
+   $ ps aux | grep '[A]pp'
 
 We will change it to run and parse this command:
 
-.. code-block:: sh
+.. code-block:: console
 
-   quota -spw
+   $ quota -spw
 
 Update test/test_command.rb
 ...........................
@@ -426,21 +426,21 @@ Steps:
 
 #. Version these changes. Click Shell button on app details view, and then commit the changes:
 
-    .. code-block:: sh
+    .. code-block:: console
 
-       git add .
-       git commit -m "update manifest for production"
+       $ git add .
+       $ git commit -m "update manifest for production"
 
        # if there is an external remote associated with this, push to that
-       git push origin master
+       $ git push origin master
 
 #. As the admin, sudo copy or clone this repo to production
 
-    .. code-block:: sh
+    .. code-block:: console
 
        # as sudo on OnDemand host:
-       cd /var/www/ood/apps/sys
-       git clone /users/PZS0562/efranz/ondemand/dev/quota
+       $ cd /var/www/ood/apps/sys
+       $ git clone /users/PZS0562/efranz/ondemand/dev/quota
 
 
 #. Reload the dashboard.
--- a/source/authentication/overview/configure-authentication.rst
+++ b/source/authentication/overview/configure-authentication.rst
@@ -54,10 +54,10 @@ under :file:`/etc/ood/config/ood_portal.yml`. For example, to add support for
 an authentication module with ``AuthType`` of ``my-auth``, you would modify the
 :file:`/etc/ood/config/ood_portal.yml` file as such:
 
-   .. code-block:: sh
+   .. code-block:: console
 
-      scl enable rh-ruby22 -- rake
-      scl enable rh-ruby22 -- rake install
+      $ scl enable rh-ruby22 -- rake
+      $ scl enable rh-ruby22 -- rake install
 
 
 .. code-block:: yaml
--- a/source/infrastructure/mod-ood-proxy/installation.rst
+++ b/source/infrastructure/mod-ood-proxy/installation.rst
@@ -5,31 +5,37 @@ Installation
 
 #. Download the latest stable release and unpack it:
 
-   .. code-block:: sh
+   Download the latest release as of this writing:
+   
+   .. code-block:: console
 
-      # Set latest release as of writing this
-      LATEST="{{ mod_ood_proxy_version }}"
+      $ LATEST="{{ mod_ood_proxy_version }}"
+      $ wget --content-disposition https://github.com/OSC/mod_ood_proxy/archive/v${LATEST}.tar.gz
 
-      # Download latest release
-      wget --content-disposition https://github.com/OSC/mod_ood_proxy/archive/v${LATEST}.tar.gz
+   Unpack archive:
+   
+   .. code-block:: console
+   
+      $ tar xzvf mod_ood_proxy-${LATEST}.tar.gz
 
-      # Unpack this archive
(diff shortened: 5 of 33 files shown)


source/_static/css/custom.css
```diff
@@ -17,3 +17,10 @@
 .wy-nav-top {
   background: #C8102E;
 }
+
+span.gp {
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  user-select: none;
+}
```
ericfranz: It might be good to namespace this so that it doesn't potentially affect anything else in the future.

If every code block highlighted with console is wrapped in a div with the class "highlight-console" maybe this is appropriate:

```css
div.highlight-console span.gp {
}
```

source/release-notes/v1.3-release-notes.rst
```diff
@@ -65,10 +66,10 @@ Upgrading from v1.2
       haven't been doing this before then now is a good time to start. Feel
       free to contact us if you have trouble with this.
 
-   .. code-block:: shell
+   .. code-block:: console
 
       # Copy over our app configuration (if they exist)
-      mkdir -p /etc/ood/config/apps
+      $ mkdir -p /etc/ood/config/apps
```
nickjer: There are commands below this.

laser505t:
@ericfranz I made the change to the CSS file that you requested. Why does it say there is still a requested change?

source/app-development/interactive/template.rst
```diff
@@ -171,9 +171,9 @@ script:
    Example
      Find a random port that is available within the default range
 
-     .. code-block:: shell
+     .. code-block:: console
 
-        port=$(find_port)
+        $ port=$(find_port)
```
nickjer: This is actually part of a shell script, so please revert this change.

source/app-development/interactive/template.rst
```diff
@@ -185,15 +185,15 @@ script:
    Example
      Generate a random password 16 characters long
 
-     .. code-block:: shell
+     .. code-block:: console
 
-        password=$(create_passwd 16)
+        $ password=$(create_passwd 16)
```
nickjer: This is actually part of a shell script, so please revert this change.

source/app-development/interactive/template.rst
```diff
@@ -185,15 +185,15 @@ script:
    Example
      Generate a random password 16 characters long
 
-     .. code-block:: shell
+     .. code-block:: console
```
nickjer: This is actually part of a shell script, so please revert this change.

source/app-development/interactive/template.rst
```diff
@@ -171,9 +171,9 @@ script:
    Example
      Find a random port that is available within the default range
 
-     .. code-block:: shell
+     .. code-block:: console
```
nickjer: This is actually part of a shell script, so please revert this change.

source/infrastructure/mod-ood-proxy/installation.rst
```diff
@@ -5,31 +5,37 @@ Installation
...
+   .. code-block:: console
 
-      # Set latest release as of writing this
-      LATEST="{{ mod_ood_proxy_version }}"
+      $ LATEST="{{ mod_ood_proxy_version }}"
+      $ wget --content-disposition https://github.com/OSC/mod_ood_proxy/archive/v${LATEST}.tar.gz
 
-      # Download latest release
-      wget --content-disposition https://github.com/OSC/mod_ood_proxy/archive/v${LATEST}.tar.gz
+   Unpack archive:
```
nickjer: In fact remove all these original comment lines and pack it into a single code-block like before.

source/infrastructure/nginx-stage/installation.rst
```diff
@@ -5,34 +5,39 @@ Installation
 
 #. Download the latest stable release and unpack it:
 
-   .. code-block:: sh
-
-      # Set latest release as of writing this
-      LATEST="{{ nginx_stage_version }}"
+   Download the latest release as of this writing:
```
nickjer: Remove all comment lines and repack the code-block.

source/infrastructure/ood-auth-map/installation.rst
```diff
@@ -5,31 +5,36 @@ Installation
 
 #. Download the latest stable release and unpack it:
 
-   .. code-block:: sh
+   Download the latest release as of this writing:
```
nickjer: Remove all comment lines and repack the code-block.

source/infrastructure/ood-portal-generator/installation.rst
```diff
@@ -5,23 +5,28 @@ Installation
 
 #. Download the latest stable release and unpack it:
 
-   .. code-block:: sh
+   Download the latest release as of this writing:
```
nickjer: Remove all the comment lines and repack the code-block.

source/release-notes/v1.3-release-notes.rst
```diff
@@ -52,10 +52,11 @@ Upgrading from v1.2
    So you will want to copy over the configuration files ahead of time for best
    results:
 
-   .. code-block:: shell
-
-      # Copy over the YAML configuration file used to generate the Apache configuration file
-      sudo cp $HOME/ood/src/ood-portal-generator/config.yml /etc/ood/config/ood_portal.yml
+   Copy over the YAML configuration file used to generate the Apache configuration file:
```
nickjer: Remove this comment line as the line above explains it well enough.

source/app-development/tutorials-passenger-apps/ps-to-quota.rst
```diff
@@ -426,21 +434,25 @@ Steps:
...
+       $ git push origin master
 
 #. As the admin, sudo copy or clone this repo to production
 
-    .. code-block:: sh
-
-       # as sudo on OnDemand host:
-       cd /var/www/ood/apps/sys
-       git clone /users/PZS0562/efranz/ondemand/dev/quota
+    As sudo on OnDemand host:
```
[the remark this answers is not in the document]
laser505t: Should ``sudo`` be added to these commands, or the prompt changed to #, or just leave it?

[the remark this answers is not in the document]
nickjer: Good point, I am not sure we should be recommending a `sudo git ...` though. I think you should replace that code line with:

```console
$ sudo cp -r /users/PZS0562/efranz/ondemand/dev/quota .
```

source/authentication/overview/configure-authentication.rst
```diff
@@ -54,12 +54,6 @@ under :file:`/etc/ood/config/ood_portal.yml`. For example, to add support for
 an authentication module with ``AuthType`` of ``my-auth``, you would modify the
 :file:`/etc/ood/config/ood_portal.yml` file as such:
 
-   .. code-block:: sh
```
MorganRodgers: @ericfranz this section seems to have been removed by Nick, and so far as I can see was not supposed to be here in the first place. I don't know the point of calling `rake` or `rake install` at this point in the documentation would be. Thoughts?

MorganRodgers: Confirmed good delete.

PR #93, opened 2018-07-31
Console to sh

MorganRodgers:
Given that the copy and paste behavior of `user-select` doesn't work reliably across browsers, we are going to prefer using the `sh` directive over `console`. We will also remove all `$` from `sh/console` blocks. Where it is necessary to distinguish between console commands and console output `#` has been used to provide different highlighting / copy-and-paste safety for output.

This PR also includes minor edits for content like PR #87 that it is replacing.

### DIFF
--- a/source/release-notes/v1.3-release-notes.rst
+++ b/source/release-notes/v1.3-release-notes.rst
@@ -64,9 +64,10 @@ Upgrading from v1.2
       haven't been doing this before then now is a good time to start. Feel
       free to contact us if you have trouble with this.
 
-   .. code-block:: sh
+  Copy over our app configuration (if they exist).
 
-      # Copy over our app configuration (if they exist)
+  .. code-block:: sh
+
       mkdir -p /etc/ood/config/apps
       [[ -e "/var/www/ood/apps/sys.bak/dashboard/.env.local" ]] \
         && sudo mkdir -p /etc/ood/config/apps/dashboard \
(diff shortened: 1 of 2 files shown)


source/release-notes/v1.3-release-notes.rst
```diff
@@ -108,45 +107,45 @@ Upgrading from v1.2
    If all went well, you should have a directory structure that looks *similar*
    to:
 
-   .. code-block:: console
+   .. code-block:: sh
 
       $ tree /etc/ood/config
```
ericfranz: Remove `$ `

source/release-notes/v1.3-release-notes.rst
```diff
@@ -65,7 +64,7 @@ Upgrading from v1.2
       haven't been doing this before then now is a good time to start. Feel
       free to contact us if you have trouble with this.
 
-   .. code-block:: shell
+   .. code-block:: sh
 
       # Copy over our app configuration (if they exist)
```
ericfranz: Instead of a comment at the top, this should be a separate line of text outside of the code block.

ericfranz:
I'd also add a css override of the theme so we don't italicize the comments, since we are now using these for displaying the output of commands.

The class that needs modified is

```css
.c1 {
    color: #998;
    font-style: italic;
}
```

Notice the difference:

c1_italicized

c1_normal

PR #107, opened 2019-01-03
Add versions at bottom

ericfranz:
Note: its preferable to auto-generate this hard-coded list. That way release branches could get this same version list added at the bottom.

To do that, we would add code like this to conf.py:

```python
# with python 2
ood_releases = subprocess.check_output("git branch -r --list 'origin/release-[0-9].[0-9]' | sed 's/origin\///g'", shell=True).decode('UTF-8').strip().split()
ood_versions = [['master', '/master/']] + list(map(lambda x: [x.replace('release-', ''), '/' + x + '/'], ood_releases)) + [['develop', '/develop/']]

# with python 3
ood_releases = subprocess.getoutput("git branch -r --list 'origin/release-[0-9].[0-9]' | sed 's/origin\///g'").strip().split()
ood_versions = [['master', '/master/']] + list(map(lambda x: [x.replace('release-', ''), '/' + x + '/'], ood_releases)) + [['develop', '/develop/']]

# which produces an array like
[
  ['master', '/master/'],
  ['1.3', '/release-1.3/'],
  ['1.2', '/release-1.2/'],
  ['1.1', '/release-1.1/'],
  ['1.0', '/release-1.0/'],
  ['develop', '/develop/']
]

# and then update html_context in conf.py:
    'conf_py_path': '/source/',
    'versions' : ood_versions
}
```

If using python 3 we would use `subprocess.getoutput`. And then we would need to update the docker image so it has git available, as well as python 3.

Then we would add these modifications to each release branch so on each new release all the branches display a link to the new release (assuming we remember to trigger a rebuild of each release branch).

### DIFF
--- a/source/_templates/versions.html
+++ b/source/_templates/versions.html
@@ -1,0 +1,26 @@
+<div class="rst-versions" data-toggle="rst-versions" role="note" aria-label="versions">
+  <span class="rst-current-version" data-toggle="rst-current-version">
+    <span class="fa fa-book"> Documentation Versions</span>
+    v: {{ current_version }}
+    <span class="fa fa-caret-down"></span>
+  </span>
+  <div class="rst-other-versions">
+    <dl>
+      <dt>{{ _('Versions') }}</dt>
+      {% for slug, url in versions %}
+        <dd><a href="https://{{ github_user }}.github.io/{{ github_repo }}/{{ url }}{{ pagename }}.html">{{ slug }}</a></dd>
+      {% endfor %}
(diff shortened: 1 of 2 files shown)


ericfranz:
We might open an issue with the above suggestion, and just merge this into develop so it is available 1.4, master, and develop branches (but not earlier).

ericfranz:
We could also do it for 1.3. Also a squash & merge will enable more easily adding to other branches (as it will be one commit).

PR #108, opened 2019-01-03
Add release notes for 1.4

ericfranz:
https://osc.github.io/ood-documentation-test/v1.4/

# TODO

- [x] Update installation/install-software.html
- [x] Show how to set `WHITELIST_PATH` in `nginx_stage`
- [x] Search for references to old dependencies
- [x] ... ?

### DIFF
--- a/source/installation/install-software.rst
+++ b/source/installation/install-software.rst
@@ -7,12 +7,10 @@ We will use `Software Collections`_ to satisfy majority of the following
 software requirements:
 
 - `Apache HTTP Server 2.4`_
-- `NGINX 1.6`_
-- `Phusion Passenger 4.0`_
-- `Ruby 2.2`_ with :command:`rake`, :command:`bundler`, and development
+- `Ruby 2.4`_ with :command:`rake`, :command:`bundler`, and development
   libraries
-- `Node.js 0.10`_
-- `Git 1.9`_
+- `Node.js 6`_
+- `Git 2.9`_
 
 .. note::
 
@@ -24,7 +22,7 @@ software requirements:
    CentOS 6/7
      .. code-block:: sh
 
-        sudo yum install centos-release-scl
+        sudo yum install epel-release centos-release-scl
 
    RHEL 6
      .. code-block:: sh
@@ -88,7 +86,7 @@ software requirements:
 .. _apache http server 2.4: https://www.softwarecollections.org/en/scls/rhscl/httpd24/
 .. _nginx 1.6: https://www.softwarecollections.org/en/scls/rhscl/nginx16/
 .. _phusion passenger 4.0: https://www.softwarecollections.org/en/scls/rhscl/rh-passenger40/
-.. _ruby 2.2: https://www.softwarecollections.org/en/scls/rhscl/rh-ruby22/
-.. _node.js 0.10: https://www.softwarecollections.org/en/scls/rhscl/nodejs010/
-.. _git 1.9: https://www.softwarecollections.org/en/scls/rhscl/git19/
+.. _ruby 2.4: https://www.softwarecollections.org/en/scls/rhscl/rh-ruby24/
+.. _node.js 6: https://www.softwarecollections.org/en/scls/rhscl/rh-nodejs6/
+.. _git 2.9: https://www.softwarecollections.org/en/scls/rhscl/rh-git29/
 .. _ohio supercomputer center: https://www.osc.edu/
--- a/source/installation/install-software.rst
+++ b/source/installation/install-software.rst
@@ -84,8 +84,6 @@ software requirements:
 
 .. _software collections: https://www.softwarecollections.org/en/
 .. _apache http server 2.4: https://www.softwarecollections.org/en/scls/rhscl/httpd24/
-.. _nginx 1.6: https://www.softwarecollections.org/en/scls/rhscl/nginx16/
-.. _phusion passenger 4.0: https://www.softwarecollections.org/en/scls/rhscl/rh-passenger40/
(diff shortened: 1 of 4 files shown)


source/release-notes/v1.4-release-notes.rst
```diff
@@ -0,0 +1,131 @@
...
+- `Experimental SGE/UGE support`_
+- `Add ability to include Javascript in Batch Connect apps`_
+
+
+Upgrading from v1.3
+-------------------
+
+To upgrade run:
+
+TODO is there a way to manage yum repos such that the dependencies are resolved without explicit installs?
```
treydock: Yes.  Upgrade will look something like this:

```
yum install -y https://yum.osc.edu/ondemand/1.4/ondemand-release-web-1.4-1.el7.noarch.rpm
yum clean all
yum update ondemand
```

source/release-notes/v1.4-release-notes.rst
```diff
@@ -0,0 +1,128 @@
...
+
+.. code-block:: bash
+
+    sudo yum install -y https://yum.osc.edu/ondemand/1.4/ondemand-release-web-1.4-1.el7.noarch.rpm
+    sudo yum clean all
+    sudo yum update ondemand
+
+    # Optionally remove dependencies from prior versions of OOD
+    yum remove rh-nodejs010\* rh-passenger40\* rh-ruby22\*
```
treydock: Also need to remove `nginx16\*`.

treydock: @ericfranz @MorganRodgers When we are ready to test 1.4 further at OSC I can run this `yum remove` command on OSC test nodes to ensure what we're advising doesn't cause issues.

source/release-notes/v1.4-release-notes.rst
```diff
@@ -0,0 +1,128 @@
...
+- `Experimental SGE/UGE support`_
+- `Add ability to include Javascript in Batch Connect apps`_
+
+
+Upgrading from v1.3
+-------------------
+
+To upgrade run:
+
+TODO is there a way to manage yum repos such that the dependencies are resolved without explicit installs?
```
treydock: Can probably remove this TODO with latest changes.

source/release-notes/v1.4-release-notes.rst
ericfranz: I don't think rh-ror41 was ever used outside of OSC.

treydock: This is what I'm running to clean up OSC dev and test systems, minor change from what's in this PR:

```
yum remove nodejs010\* rh-passenger40\* rh-ruby22\* nginx16\*
```

For OSC systems it turns out the above will also remove `rh-ror41`.

source/installation/install-software.rst
```diff
@@ -88,7 +86,7 @@ software requirements:
 .. _apache http server 2.4: https://www.softwarecollections.org/en/scls/rhscl/httpd24/
 .. _nginx 1.6: https://www.softwarecollections.org/en/scls/rhscl/nginx16/
```
treydock: Need to remove this reference.

source/installation/install-software.rst
```diff
@@ -88,7 +86,7 @@ software requirements:
...
 .. _phusion passenger 4.0: https://www.softwarecollections.org/en/scls/rhscl/rh-passenger40/
```
treydock: Need to remove this reference.

source/release-notes/v1.4-release-notes.rst
```diff
@@ -0,0 +1,136 @@
...
+it was updated from for each of the system web applications in this release.
+
+
+Details
+-------
+
+Upgrade to Ruby 2.4, NodeJS 6, Passenger 5
+............................................
+
+This upgrade updates our dependencies to Software Collections Ruby 2.4 and NodeJS 6. Passenger is also seeing an upgrade to version 5, but until Passenger 5 is supported by SCL OSC will host the Passenger 5 and NGINX 1.14 RPMs which are built based on "@efranz where did we get the Passenger RPM build information from?".
```
treydock: We use this repo for the RPM builds: https://github.com/phusion/passenger_rpm_automation.

treydock:
Do we need mention of SLURM 18 support?

ericfranz:
Missing from this list of highlights:

- Customizable error pages for missing home directory during the first login flow for sites using pam_mkhomedir.so
- Introduction of a whitelist mode for apps which is disabled by default. This change means that by default, when deploying a new app, if properly configured it will appear in the Dashboard's navigation menu without the need for changing configuration. https://github.com/OSC/ood-dashboard/issues/295
- Security Enhancement - each PUN now has its own unique per-user secret key base string which is generated at first launch if it doesn't exist. This is used by Rails apps to encrypt cookies and overrides the default one set in https://github.com/OSC/ood-dashboard/blob/a20cae91fb0f636c8a7e3fc8dfe508d9afa34d50/.env.production#L1


Other:

- for "Better environment setting" its `pun_custom_env` not `custom_env` and you can link to the example here: https://github.com/OSC/ondemand/blob/d85a3982d69746144d12bb808d2419b42ccc97a1/nginx_stage/share/nginx_stage_example.yml#L26-L43

ericfranz:
For the pam_mkhomedir.so feature it is probably sufficient for now to link to the Discourse discussion https://discourse.osc.edu/t/launching-ondemand-when-home-directory-does-not-exist/53/ which has screenshots and details.

In fact, we could do this as a temporary solution to quickly document some of these things.

MorganRodgers:
I think this is ready to merge.

PR #109, opened 2019-01-04
[WIP] Reorg

ericfranz:
See the example: https://osc.github.io/ood-documentation-test/reorg/

Enhancements include:

1. Lighter navbar text is easier to read
2. New sections Install, Configure, Customize, Extend

TODO:

1. Add in documentation for missing sections
2. Add missing sections (quotas and other features)


ericfranz:
Actually, I think we should flatten the left customization menu:

- Announcements
- MOTD
- Branding
- Default SSH Host
- Files Shortcuts
- Custom Job Composer Templates

Move to the Configure section

- Whitelist directories

And maybe the difference between Configure and Customize (that is in my head) is not pronounced enough to make sense to other users.

My intention was Configure would deal with cluster config, nginx_stage.yml, ood-portal.yml, and things like handling first user login flow when home directory doesn't exist with custom error pages. The custom error pages seems more like a "customization" though.

Maybe a list of 20 items is not overwhelming...

ericfranz:
Well the hierarchy does help if you want to know "I want to customize Active Jobs. What are all my options?"

PR #113, opened 2019-01-04
[WIP] - The Customization Page

MorganRodgers:
Adding PR as a checklist and place for discussion.

- [x] Setting PUN environment using `nginx_stage.
- [x] How to add announcements / MOTD
- [x] Branding ... ?
- [x] Setting a Title
- [x] Adding Logos
- [x] Setting Colors
- [ ] Setting a directory whitelist for the JobComposer, File Editor and File Explorer
- [ ] Setting default SSH host
- [ ] Adding templates for the Job Composer
- [ ] ... other stuff?

### DIFF
--- a/source/customization.rst
+++ b/source/customization.rst
@@ -3,10 +3,26 @@
 Customization
 =============
 
-Announcements
----------------
+Or the many OOD environment variables:
 
-.. attention:: TODO
+Setting the PUN Environment With ``nginx_stage.yml``
+----------------------------------------------------
+
+The configuration file ``nginx_stage.yml`` is the best way to set the environment which the PUN will use. Defining the mapping ``pun_custom_env`` allows setting Open OnDemand specific variables and is used for branding, message of the day setting, quota warnings, whitelists and other uses. For environment variables that OOD applications may need, are already in the environment and should be exposed to OOD (``$PATH``, ``LD_LIBRARY_PATH``, etc...) define the sequence ``pun_custom_env_declarations``.
+
+An example of both of these uses may be found in `nginx_stage_example.yml <https://github.com/OSC/ondemand/blob/d85a3982d69746144d12bb808d2419b42ccc97a1/nginx_stage/share/nginx_stage_example.yml#L26-L43>`__
+
+Announcements and Message of the Day (MOTD)
+-------------------------------------------
+
+Two hooks for customizing the Dashboard are Announcements and MOTD. Announcements get the classes ``alert alert-warning`` and appear above the ``OOD_DASHBOARD_LOGO`` . Announcement files are expected to be found at: ``/etc/ood/config/announcement.(md|yml)`` or ``/etc/ood/config/announcements.d/any_file_name.(md|yml)``. To display a MOTD file on the Dashboard ensure that the environment variables  ``MOTD_PATH`` and ``MOTD_FORMAT`` are set like so:
+
+   .. code-block:: sh
+
+      MOTD_PATH="/path/to/your/motd" # this supports both file and RSS feed URIs
+      MOTD_FORMAT="$MOTD_FORMAT_HERE" # markdown, txt, rss
+
+.. note:: Warnings about the announcement file being missing may be present in users' nginx logs. Despite the warning the Dashboard will still function normally without those files being present.
 
 Branding
 -------------------
--- a/source/customization.rst
+++ b/source/customization.rst
@@ -1,21 +1,19 @@
 .. _customization:
 
-Customization
-=============
-
-Or the many OOD environment variables:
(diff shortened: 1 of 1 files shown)


source/customization.rst
```diff
@@ -1,32 +1,49 @@
...
+-------------------------------------------
+
+Two hooks for customizing the Dashboard are Announcements and MOTD. Announcements get the classes ``alert alert-warning`` and appear above the ``$OOD_DASHBOARD_LOGO`` . Announcement files are expected to be found at: ``/etc/ood/config/announcement.(md|yml)`` or ``/etc/ood/config/announcements.d/any_file_name.(md|yml)``. To display a MOTD file on the Dashboard ensure that the environment variables ``$MOTD_PATH`` and ``$MOTD_FORMAT`` are set like so:
+
+   .. code-block:: sh
+
+      MOTD_PATH="/path/to/your/motd" # this supports both file and RSS feed URIs
+      MOTD_FORMAT="$MOTD_FORMAT_HERE" # markdown, txt, rss
+
+.. note:: Warnings about the announcement file being missing may be present in users' nginx logs. Despite the warning the Dashboard will still function normally without those files being present.
```
ericfranz: We should open an issue and fix this in 1.5. This is the fault of code trying to open the file and catching an exception instead of first checking if the file exists and can be read.

source/customization.rst
```diff
@@ -1,32 +1,49 @@
...
+Customization (or the many OOD environment variables)
+=====================================================
 
-Announcements
----------------
+Setting the PUN Environment With ``nginx_stage.yml``
+----------------------------------------------------
 
-.. attention:: TODO
+The configuration file ``nginx_stage.yml`` is the best way to set the environment which the PUN will use. Defining the mapping ``pun_custom_env`` allows setting Open OnDemand specific variables and is used for branding, message of the day setting, quota warnings, whitelists and other uses. For environment variables that OOD applications may need, are already in the environment and should be exposed to OOD (``$PATH``, ``$LD_LIBRARY_PATH``, etc...) define the sequence ``pun_custom_env_declarations``.
```
ericfranz: It is still valuable at this stage to be able to set some per-app configuration in the individual app's dotenv files. In fact I think most of the configuration variables at this stage should be still set in the apps themselves and not in nginx_stage.yml. I don't think it is appropriate for message of the day or quota warnings. I think it is appropriate for the whitelist and branding since these are intended to be used across all apps. The motd I would keep with the dashboard config for now. I think its actually a negative that once the PUN has it, 100% of the apps have it set, regardless if whether you are running an app in production mode, or development mode, whether it is a core app or a third party app. And default dotenv functionality would not override it. So I think we be careful what we encourage to be set here.

This is at least what Jeremy and I have discussed in the passed. Though feel free to push back and we can discuss. In the future I would imagine we would design a better centralized OnDemand configuration that all apps would have access to, that would not be in the environment. And branding might be delegated to a css file.

source/customization.rst
```diff
@@ -1,32 +1,49 @@
...
+
+An example of both of these uses may be found in `nginx_stage_example.yml <https://github.com/OSC/ondemand/blob/d85a3982d69746144d12bb808d2419b42ccc97a1/nginx_stage/share/nginx_stage_example.yml#L26-L43>`__
+
+Announcements and Message of the Day (MOTD)
+-------------------------------------------
```
ericfranz: I would split these into separate sections as they really are separate features that aren't related to each other at all.

source/customization.rst
```diff
@@ -1,32 +1,49 @@
...
+
+Two hooks for customizing the Dashboard are Announcements and MOTD. Announcements get the classes ``alert alert-warning`` and appear above the ``$OOD_DASHBOARD_LOGO`` . Announcement files are expected to be found at: ``/etc/ood/config/announcement.(md|yml)`` or ``/etc/ood/config/announcements.d/any_file_name.(md|yml)``. To display a MOTD file on the Dashboard ensure that the environment variables ``$MOTD_PATH`` and ``$MOTD_FORMAT`` are set like so:
```
ericfranz: >appear above the ``$OOD_DASHBOARD_LOGO``

Maybe just "appear at the top of the page" - becuas announcements appear on the index page and also on the interactive app web form pages and the list of current interactive sessions and the developer views etc.

ericfranz:
See https://github.com/OSC/ood-dashboard/wiki/Configuration-and-Branding. Some of this is old (the colors for example and especially BOOTSTRAP_ stuff). 

I also wonder how much of this is so expensive to add to Sphinx that is preferable to link to Discourse or GitHub wiki till we have a better solution. Also, I've also wondered about whether we should just start doing some screencasts.

PR #121, opened 2019-01-07
An example of a bin_override wrapping sbatch

MorganRodgers:


source/installation/resource-manager/bin-override-example.rst
```diff
@@ -0,0 +1,167 @@
+.. _bin_overrides:
+
+A Working Example Of A ``bin_overrides`` Script
+===============================================
+
+``sbatch`` is the Slurm scheduler's submission client. In this example we will show how cluster configuration's ``bin_override`` may be used to change default behavior.
```
ericfranz: bin_override should be bin_overrides

ericfranz:
Looks good to me except for missing the plural on`bin_overrides` in a few locations, assuming you tested the execution of of the override. Looks promising!

PR #141, opened 2019-01-31
Update Keycloak docs for Keycloak 4

### LINKED ISSUE #96: Update Keycloak documentation for Keycloak 4.x
There are several places in our Keycloak documentation that are no longer valid with Keycloak 4.x.  Mainly is there is no longer a concept of client templates, they are now client scopes and assigned to clients either specifically or defaulted based on realm defaults.  The default scopes for OpenID Connect are good enough that our docs won't need to create any.

treydock:
Will resolve #96


treydock:
Tested on webdev07 like before.  It's still running if want to verify.

PR #142, opened 2019-02-01
RPM will move /var/lib/nginx/config to /var/lib/nginx/config.rpmsave

treydock:


treydock:
Don't merge this yet, I am going to have to make changes to RPM and will make the documentation changes here.

treydock:
Added commit to account for recent RPM changes, this is good now.

ericfranz:
Okay I merged the other so we now always source the default profile first, so this can be reworded.

treydock:
Added commit to describe the change in behavior.  I got writer's block so am open to suggestions on how to better word the warning.

PR #145, opened 2019-02-01
Thanks

MorganRodgers:
Added two additional sites that helped field test the SGE adapter.

### DIFF
--- a/source/index.rst
+++ b/source/index.rst
@@ -28,6 +28,8 @@ Special Thanks
 
 These are institutions who were early adopters or provided HPC resources for development and testing of the resource manager adapters currently available. The success of this project would not be possible without their generous support (in alphabetical order):
 
+- Bowdoin College. For early adoption and feedback with the SGE adapter.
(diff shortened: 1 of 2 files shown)


source/index.rst
```diff
@@ -28,6 +28,8 @@ Special Thanks
 
 These are institutions who were early adopters or provided HPC resources for development and testing of the resource manager adapters currently available. The success of this project would not be possible without their generous support (in alphabetical order):
 
+- Bowdoin College. For early adoption and feedback with the SGE adapter.
+- Oregon State University. For early adoption and feedback with the SGE adapter.
```
ericfranz: I was considering "early adoption" to be "early adoption of Open OnDemand" as a platform i.e. the time before we had an rpm and had zero installs. I wanted to call out the first few institutions who dived in when they knew they were the first ones to use it.

MorganRodgers: ah...up to you. I'm not too worried about the wording.

PR #147, opened 2019-02-13
Explains advantage of using bind mount with Singularity for RStudio

MorganRodgers:


ericfranz:
Comments on the entire section:

1. Rename "Add A RStudio App" to "Add an RStudio App"
2. Put the tutorial in the menu right after the "Add a Jupyter App" Since you state:

   >In a previous example we showed how an interactive Jupyter application could be set up.

3. You can add to toctree `:numbered: 1` in some place (I don't remember where) that will add the numbering to each section of the tutorial, which is helpful. There might also be a configuration option on the page itself (at the top?).

4. Maybe replace "Build The Singularity Image" section with "Get Singularity Image" and then two sub sections - use prebuilt one from Singularity hub with a link, or build your own.

5. For the `Complete OSC-Specific Example` section, it would be great to summarize the steps to take if someone like PSC would like to copy our app and use it on their systems. For example, step 1 might be to get the singularity image and then deploy it to the path that we use (or another path, but remember) and then add specific modules that set certain variables, and update the form.yml appropriately, and perhaps the submit.yml for their particular resource manager? Is there anything else?

ericfranz:
In fact, Software Requirements to Running RStudio seems to be an overview and the OSC example would be a tutorial or walkthrough.

PR #159, opened 2019-05-22
 Document Dependencies

kinanAlAttar:
I have edited source/index.rst to document all the system dependencies of OnDemand that I have found.

### DIFF
--- a/source/index.rst
+++ b/source/index.rst
@@ -89,6 +89,31 @@ These are institutions who were early adopters or provided HPC resources for dev
    applications
    user-documentation
 
+OnDemand System Dependencies
+----------------------------
+
+The following table lists the system dependencies of OnDemand while it is running.  
+
+
+.. list-table:: System dependencies
+   :header-rows: 1
+   :stub-columns: 1
+
+   * - Program
+     - Location Of Execution 
+   * - Rsync
+     - /var/www/ood/apps/sys/myjobs/app/models/workflow.rb
+   * - GNU Timeout
+     - /var/www/ood/apps/sys/myjobs/app/models/filesystem.rb
+   * - File
+     - /var/www/ood/apps/sys/file-editor/app/controllers/pages_controller.rb
+   * - Xmllint
+     - /opt/ood/ondemand/root/usr/sbin/passenger-status
+   * - GNU Wget
+     - /opt/ood/mod_ood_proxy/lib/analytics.lua
+
+
+**Note:** GCC is a builtin dependency and will be required for installing OnDemand from Source.
 
 .. _website: https://osc.github.io/Open-OnDemand/
 .. _bowdoin: https://www.bowdoin.edu/it/resources/high-performance-computing.html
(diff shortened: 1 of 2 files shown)


source/index.rst
```diff
@@ -89,6 +89,31 @@ These are institutions who were early adopters or provided HPC resources for dev
    applications
    user-documentation
 
+OnDemand System Dependencies
```
MorganRodgers: This section should really go in `installation/install-software.rst`.

source/index.rst
```diff
@@ -89,6 +89,31 @@ These are institutions who were early adopters or provided HPC resources for dev
...
+
+The following table lists the system dependencies of OnDemand while it is running.  
+
+
+.. list-table:: System dependencies
+   :header-rows: 1
+   :stub-columns: 1
+
+   * - Program
+     - Location Of Execution
```
MorganRodgers: Instead of 'Location of Execution' let's call it 'OnDemand Component'.

source/index.rst
```diff
@@ -89,6 +89,31 @@ These are institutions who were early adopters or provided HPC resources for dev
...
+     - /var/www/ood/apps/sys/myjobs/app/models/filesystem.rb
+   * - File
+     - /var/www/ood/apps/sys/file-editor/app/controllers/pages_controller.rb
+   * - Xmllint
+     - /opt/ood/ondemand/root/usr/sbin/passenger-status
+   * - GNU Wget
+     - /opt/ood/mod_ood_proxy/lib/analytics.lua
+
+
+**Note:** GCC is a builtin dependency and will be required for installing OnDemand from Source.
```
MorganRodgers: 'builtin' should be 'build time', so this should be: 

```GCC is a build time dependency and is only required for installing OnDemand from source.```

File: source/index.rst [hunk 4]
```diff
@@ -89,6 +89,31 @@ These are institutions who were early adopters or provided HPC resources for dev
    applications
    user-documentation
 
+OnDemand System Dependencies
+----------------------------
+
+The following table lists the system dependencies of OnDemand while it is running.
```
MorganRodgers: I'd replace this line with:

```
For users attempting to install OnDemand from source on non-CentOS/RHEL systems it may be useful to know that OnDemand has several other run and build time dependencies that may need to be installed.
```

source/index.rst
kinanAlAttar: I have resolved this along with the other requests of change.

source/index.rst
```diff
@@ -89,7 +89,6 @@ These are institutions who were early adopters or provided HPC resources for dev
    applications
    user-documentation
 
-
```
MorganRodgers: Note another unnecessary white space change. Now, these aren't a huge deal, but it's better not to change things that we don't need to. So how could this change be undone? One way to return a file to the state that it was in at a specific commit or tag is to check it out.

```sh
git checkout develop -- source/index.rst
```

In this case you do not need to make that change, but knowing how to get a specific version of a file in Git is useful.

kinanAlAttar: I will keep that in mind. Thanks.

PR #160, opened 2019-05-23
New release notes

MorganRodgers:


ericfranz:
Missing:

- Application and Infrastructure version changes sections: https://osc.github.io/ood-documentation/master/release-notes/v1.5-release-notes.html#application-version-changes (for example https://github.com/osc/ood-dashboard/compare/v1.31.0...v1.33.4)
-  Support fetching quota from a URL (thank you Dan Dietz from Purdue U.) - I missed adding this to the CHANGELOG file

It is worth building the application and infrastructure change sections in the documentation, with the nice GitHub diff links, and then reviewing the actual code changes to see if there is anything notable that we missed adding to the CHANGELOG to call out in the release notes.

These two need new documentation added:

- Support fetching quota from a URL (how to enable, how it works) - probably in https://osc.github.io/ood-documentation/master/customization.html#disk-quota-warnings-on-dashboard with an explanation that this does an additional URL request during the dashboard request handling, so could be slower
- New localization options, how to see the list of what can be changed (by inspecting the YAML file) and what is important is at the top; and maybe some common modifications (like the front page, etc.). In the quota files section of customization, explaining how to change the message text might be useful (so if you don't update every 5 minutes, for example).

source/release-notes/v1.6-release-notes.rst
```diff
@@ -0,0 +1,178 @@
...
+
+Added rendering of HTML or Markdown in Job Composer template manifests
+......................................................................
+
+If the environment variable ``RENDER_TEMPLATE_NOTES_AS_MARKDOWN`` is set to a truthy value then job template manifests displayed in the Job Composer that are written in Markdown will be run through the RedCarpet renderer. HTML values will always be rendered as HTML. This will allow sites to embed images or links to documentation in their templates. (`ood-mjobs#291`_)
+
+.. _ood-mjobs#291: https://github.com/OSC/ood-myjobs/issues/278
+
+
+Added L18n hooks to the Dashboard and Job Composer
```
ericfranz: [it is "I18n" and "L10n"](https://blog.mozilla.org/l10n/2011/12/14/i18n-vs-l10n-whats-the-diff/) not L18n.

But this is also technically incorrect.  "Added localization options to the Dashboard and Job Composer" might be better. The problem with the nomenclature is there is no single noun that represents the "feature we added". Rather, two verbs, one for adding the feature (internationalizing the app by abstracting locale-specific strings) and two for using the feature (localizing the app to your site by providing the translations).

source/release-notes/v1.6-release-notes.rst
```diff
@@ -0,0 +1,178 @@
...
+==================
+
+Highlights in 1.6:
+
+- `OnDemand passed a security audit!`_
+- `Added job array support for LSF and PBSPro adapters`_
+- `Added L18n hooks to the Dashboard and Job Composer`_
+- `Added rendering of HTML or Markdown in Job Composer template manifests`_
+- `Changed ngnix_stage to always remove stale PID files`_
+- `Fixed Active Jobs to show empty cell if account is nil`_
```
ericfranz: Instead of having separate sections for each bug fix and each adapter improvement lets combine these into two sections:

1. Fixed several bugs
2. Improved SGE and Slurm adapters

The content of the original sections can be added to each, as separate paragraphs, or in the case of the bug fixes, as bulleted list

source/release-notes/v1.6-release-notes.rst
```diff
@@ -0,0 +1,178 @@
...
+- `Fixed crash in nginx_stage`_
+- `Fixed Job Composer layout when job name is long`_
+- `Fixed Time Used format for Active Jobs`_
+- `Fixed warning about eager loading gems`_
+- `Improved Grid Engine adapter`_
+- `Improved Slurm adapter`_
+- `Updated web-application dependencies`_
```
ericfranz: We update dependencies every time so it doesn't need to be explicitly called out, unless its significant dependency like Passenger version or default Ruby version.

ericfranz:
When documenting the localization options, we should also document how to change the locale if that is desirable.

source/release-notes/v1.6-release-notes.rst
```diff
@@ -0,0 +1,178 @@
...
+
+   .. code-block:: sh
+
+       sudo yum clean all
+       sudo yum update ondemand
+
+Details
+-------
+
+OnDemand passed a security audit!
```
ericfranz: remove the "!" :-)

PR #162, opened 2019-05-24
Update build deps

MorganRodgers:
Quoting from the PR I made on the other repo: https://github.com/OSC/docker-sphinx/pull/1:

Update to Python3 because of the impending Python2 EOL. Also switch to using Pipenv to manage our dependencies because the ood-documentation repo uses it, and we need specific versions of a few things because of build errors and CVEs.

The 'lexer error' I reference happens because we are running sphinx-build with the -W flag which breaks the build if there is a warning. Newer versions of the JSON highlighter complain when attempting to highlight invalid JSON in https://github.com/OSC/ood-documentation/blob/develop/source/customization.rst, thus breaking the build. By using specific versions of pygmentize and sphinx we won't see that error.


MorganRodgers:
Closing because I accidentally included irrelevant changes.

PR #167, opened 2019-05-28
Modify System Security

Mreyna3:
Modified the 1.5 build instructions, to display more in depth detail surrounding : 

- Disabling SELinux 
-Opening iptables 

Instructions pertain to centOS/RHEL 6/7

### DIFF
--- a/build/html/installation/modify-system-security.html
+++ b/build/html/installation/modify-system-security.html
@@ -1,0 +1,374 @@
+
+
+<!DOCTYPE html>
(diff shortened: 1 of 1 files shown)


MorganRodgers:
@Mreyna3 we are intentionally [ignoring the build directory](https://github.com/OSC/ood-documentation/blob/develop/.gitignore#L18). Please only commit changes to the source directory.

You can either remove the files you committed from the branch or just start a new one. Up to you.

PR #169, opened 2019-05-30
More in depth configuration of SElinux and iptables for 1.5

Mreyna3:
Modified the 1.5 build instructions, to display more in depth detail surrounding :

Disabling SELinux
-Opening iptables
Instructions pertain to centOS/RHEL 6/7

### DIFF
--- a/source/installation/modify-system-security.rst
+++ b/source/installation/modify-system-security.rst
@@ -3,25 +3,16 @@
 Modify System Security
 ======================
 
-#. Disable `SELinux`_ on the Open OnDemand host you intend on running the web
-   server on.
-
-   .. note::
-
-      If SELinux is not disabled on the web server host machine this can lead
-      to permission errors when running the web applications. SELinux may also
-      need to be disabled on the web server if you want to take advantage of
-      certain software in your web applications such as being able to access a
-      Lustre file system.
-
+#. Check `SELinux`_ status, by default SELinux is enabled on CentOS and most other RHEL-based systems.
+   
    RHEL/CentOS 6/7
       .. code-block:: sh
 
-         sestatus
+         $ sestatus
 
       .. code-block:: sh
 
-         #Output
+         #Output displays the starting status of SELinux
          **SELinux status**:             **enabled**
          SELinuxfs mount:                /sys/fs/selinux
          SELinux root directory:         /etc/selinux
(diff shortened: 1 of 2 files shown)


source/installation/modify-system-security.rst
```diff
@@ -14,6 +14,46 @@ Modify System Security
...
+         Mode from config file:          enforcing
+         Policy MLS status:              enabled
+         Policy deny_unknown status:     allowed
+         Max kernel policy version:      31
+
+
+   RHEL/CentOS 6/7
+      .. code-block:: sh
+
+         $ sudo chmod -R 757 /etc/selinux/config
```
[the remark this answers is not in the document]
Mreyna3: I removed the block you suggested. And added setenforce 0 as method for disabling the SELinux.

source/installation/modify-system-security.rst
```diff
@@ -32,27 +23,52 @@ Modify System Security
...
+      If SELinux is not disabled on the web server host machine this can lead
+      to permission errors when running the web applications. SELinux may also
+      need to be disabled on the web server if you want to take advantage of
+      certain software in your web applications such as being able to access a
+      Lustre file system.
 
-         $ sudo chmod -R 757 /etc/selinux/config  
 
-   Using your favorite text editor open the /etc/selinux/config file and set the SELINUX mode to disabled 
+   **Option 1: Disable SELinux Temporarily**
```
MorganRodgers: Using `setenforce 0` should not be presented as a separate option from editing `/etc/selinux/config`. They should be used together.

source/installation/modify-system-security.rst
```diff
@@ -32,27 +23,52 @@ Modify System Security
...
+               #       enforcing - SELinux security policy is enforced.
+               #       permissive - SELinux prints warnings instead of enforcing.
+               #       disabled - No SELinux policy is loaded.
+               SELINUX=disabled
+               # SELINUXTYPE= can take one of these two values:
+               #       targeted - Targeted processes are protected,
+               #       mls - Multi Level Security protection.
+               SELINUXTYPE=targeted
+
+      #. Reboot the OS to save the changes
```
MorganRodgers: Rebooting is not necessary to disable SELinux if the user runs `setenforce 0`.

source/installation/modify-system-security.rst
```diff
@@ -73,7 +89,7 @@ Modify System Security
 
    RHEL/CentOS 6
       .. code-block:: sh
-
```
MorganRodgers: In general you should avoid whitespace changes that are not relevant to the task at hand to keep the diffs clean. In this case you remove a tab which we did not need, so no harm done.

ericfranz:
Replaced by https://github.com/OSC/ood-documentation/pull/217

PR #178, opened 2019-06-26
Add visualizations

johrstrom:
This adds 2 architecture diagrams and a sequence diagram. I'm not super happy with the architectures, the only alternative I can think is using draw.io and saving the .drawio file (xml) *and* the image (png or similar) both. It won't be automatic in the build, but it will be nicer architecture diagrams.

https://osc.github.io/ood-documentation-test/add-visualizations/architecture.html


ericfranz:
1. Can you share on this PR a quick short example of a drawio XML and corresponding png you might consider including?
2. Regarding the container context diagram - it might be useful under the diagram to add a note about what "container" means - it is of course C4 model nomenclature and not the Singularity or Docker container.
3. My understanding of a C4 container:
   >A container is essentially a context or boundary inside which some code is executed or some data is stored. And each container is a separately deployable/runnable thing or runtime environment, typically (but not always) running in its own process space. Because of this, communication between containers typically takes the form of an inter-process communication. ". 

   In this regard, its important to know that httpd communicates with the per-user NGINX using inter-process communication, and NGINX communicates with Passenger processes using inter-process communication. So in the diagram as it exists is incorrect because the Passenger box contains the NGINX box, but really the nginx is a box, and then Passenger is another box, and that Passenger box you could say contains files app, dashboard app, ssh app, and the dashboard app loads interactive app plugins from the network file system.

4. Along with the C4 architecture diagrams it might be useful to have at the top of the page this overview diagram: image
5. Maybe we should also link to relevant presentations that cover some of the architecture in more detail. Not sure.

ericfranz:
Finally in addition to the more detailed C4 diagrams, we might move these diagrams to this location as well (from the legacy docs section):

- https://osc.github.io/ood-documentation-test/add-visualizations/applications/dashboard.html#how-it-works
- https://osc.github.io/ood-documentation-test/add-visualizations/applications/shell.html#how-it-works

etc.

johrstrom:
>     1. Can you share on this PR a quick short example of a drawio XML and corresponding png you might consider including?

I haven't developed anything in it yet, it was just an idea after I'd done all the work in PlantUML. But the diagrams/pngs would look nicer I think.  Think Visio or similar professional looking diagrams, which I think is for sure needed over what's given here.  The cooresponding XML's content is a lot of hashing, but the value of having the XML is reproducability - albeit manually. Added to the value of it being online and free, it doesn't restrict us to a given platform. 


>     2. Regarding the container context diagram - it might be useful under the diagram to add a note about what "container" means - it is of course C4 model nomenclature and not the Singularity or Docker container.
Agreed.

> 
>     3. My understanding of a C4 container:
>        > A container is essentially a context or boundary inside which some code is executed or some data is stored. And each container is a separately deployable/runnable thing or runtime environment, typically (but not always) running in its own process space. Because of this, communication between containers typically takes the form of an inter-process communication. ".
>        
>        
>        In this regard, its important to know that httpd communicates with the per-user NGINX using inter-process communication, and NGINX communicates with Passenger processes using inter-process communication. So in the diagram as it exists is incorrect because the Passenger box contains the NGINX box, but really the nginx is a box, and then Passenger is another box, and that Passenger box you could say contains files app, dashboard app, ssh app, and the dashboard app loads interactive app plugins from the network file system.
Yea, I think for sure we need a better diagramming tool. PlantUml/Graphviz just doesn't have the flexibility, or frankly the visual pleasantness that the task requires. I can come up with a quick 'C4 container' view and add it here to the PR.

> 
>     4. Along with the C4 architecture diagrams it might be useful to have at the top of the page this overview diagram: image
OK
> 
>     5. Maybe we should also link to relevant presentations that cover some of the architecture in more detail. Not sure.
I added a "General" tab to this, maybe under it have a page that has links to all presentations. Then link the two?

ericfranz:
I'm new to draw.io but what is interesting to me is their repos: https://github.com/jgraph/drawio

>draw.io uses the mxGraph library as the base of the stack

- https://github.com/jgraph/mxgraph
- https://github.com/jgraph/drawio-desktop is an Electron version of the drawio app

So call me crazy, but it would seem that we would be able to:

1. use the mxgraph library to auto-generate png from the xml that drawio works with, during the sphinx build
2. open the xml files using the electron version which could be included in the docker image used to build the app (or just included as a separate app?)

Just crazy thinking right now. I don't know everything involved; but maybe we could have an all in one environment for editing these files? I'm still a fan of the simplicity of violetuml but thats Java and I don't think it is maintained.

ericfranz:
>I added a "General" tab to this, maybe under it have a page that has links to all presentations. Then link the two?

I don't necessarily want to replicate what is already on http://openondemand.org/. There we have tables for all the webinars, conferences, publications, etc.

I do think, however, if you are explicitly wanting to better understand the architecture of OnDemand these diagrams are the start, and then you kinda want to know "where to learn more". So maybe in that case, at the bottom of this Architecture section we would have a "where to learn more" and there we could provide suggestions. One might be (see these particular publications or webinars etc.); another might be (install Vagrant image and inspect the pieces yourself, or look at the infrastructure repo and packaging repo).

johrstrom:
ood_container_view

Here is, what I hope, you think is a better container view made from draw.io.  What are your thoughts? I know I need to have descriptions on the connecting lines but I thought I'd get some feedback to see if it's even the right direction.

ericfranz:
That's an improvement and fixes the problems with the previous. Also the blue is less harsh than the red which is nice. I'd put in the labels and call it done. Does the XML indicate that it was generated from draw.io just so we don't forget?

johrstrom:
Awesome. Yea I'd used red because all the OSU/OSC red that I've been noticing we tend to use. I'll fix this up and commit it. The xml does seem to indicate the host (www.draw.io), but I was actually going to put a README.md in the architecture directory along with the images so we don't forget.

PR #180, opened 2019-07-08
[ON HOLD] add pam authentication docs

johrstrom:
Docs added for PAM authentication based off of community feedback here:

https://discourse.osc.edu/t/can-ood-auth-be-handled-by-pam/81/4

### DIFF
--- a/source/authentication/pam.rst
+++ b/source/authentication/pam.rst
@@ -73,4 +73,4 @@ community provided, so if you have issues with them, let us know!
 
       sudo service httpd24-httpd condrestart
 
-#. Cleanup the files extracted in step 4 and disable epel if you wish.
+#. Cleanup the files extracted in step 4.
--- a/source/authentication/pam.rst
+++ b/source/authentication/pam.rst
@@ -1,0 +1,76 @@
+.. _authentication-tutorial-pam-authnz:
+
+PAM Authentication using mod Authnz
+=======================================
+
+These are the steps you'll need to take to enable PAM authentication in
+OnDemand using the ``mod_authnz_external`` httpd plugin.  These steps are
+community provided, so if you have issues with them, let us know!
+
+
+#. Enable epel if required
+
+    .. code-block:: sh
+
+       yum install epel-release
+
+#. Install pwauth
+
+    .. code-block:: sh
+
+       yum install pwauth
+
+
+#. Download the mod_authnz_external rpm
+
+    .. code-block:: sh
+
+       wget http://download.fedoraproject.org/pub/epel/7/x86_64/Packages/m/mod_authnz_external-3.3.1-7.el7.x86_64.rpm
+
+
+#. Unpack the files from the rpm.
+
+    .. code-block:: sh
+
+       rpm2cpio mod_authnz_external-3.3.1-7.el7.x86_64.rpm | cpio -idmv
+
+
+#. Copy the unpacked files to the appropriate httpd24 locations.
+
+    .. code-block:: sh
+
+       cp usr/lib64/httpd/modules/mod_authnz_external.so /opt/rh/httpd24/root/usr/lib64/httpd/modules
+       cp etc/httpd/conf.d/authnz_external.conf /opt/rh/httpd24/root/etc/httpd/conf.d
+
+#. Modify /etc/ood/config/ood_portal.yml to have these yaml elements.
+
+    .. warning::
+       Be careful of copy/paste errors with indentation and ' and " characters.
+
+    .. code-block:: yaml
+
+       - 'DefineExternalAuth pwauth pipe /usr/bin/pwauth'
+
+       auth:
+       - 'SSLRequireSSL'
+       - 'AuthType Basic'
+       - 'AuthName "private"'
+       - 'AuthBasicProvider external'
+       - 'AuthExternal pwauth'
+       - 'RequestHeader unset Authorization'
+       - 'require valid-user'
+
+
+#. Generate config files with the ood_portal command.
+
+    .. code-block:: sh
+
+      /opt/ood/ood-portal-generator/sbin/update_ood_portal
+
+#. Restart apache httpd
+
+    .. code-block:: sh
+
+      sudo service httpd24-httpd condrestart
+
+#. Cleanup the files extracted in step 4 and disable epel if you wish.
--- a/source/authentication.rst
+++ b/source/authentication.rst
@@ -23,4 +23,5 @@ OpenID Connect with KeyCloak).
    authentication/overview
    authentication/tutorial-oidc-keycloak-rhel7
    authentication/duo-2fa-with-keycloak
+   authentication/pam
 


treydock:
The page for this should mention it's what's necessary to do PAM on RHEL/CentOS 6. The steps for RHEL/CentOS 7 I would hope involve "yum install mod_authnz_external" and not downloading the RPM and unpacking it's contents.

Also soon as EPEL releases a new version of mod_authnz_external that wget will fail because EPEL does not keep around old versions.

source/authentication/pam.rst
```diff
@@ -0,0 +1,76 @@
...
+
+PAM Authentication using mod Authnz
+=======================================
+
+These are the steps you'll need to take to enable PAM authentication in
+OnDemand using the ``mod_authnz_external`` httpd plugin.  These steps are
+community provided, so if you have issues with them, let us know!
+
+
+#. Enable epel if required
```
treydock: This is required for RHEL/CentOS 7 only.

treydock: Actually nevermind, pwauth comes from EPEL.

source/authentication/pam.rst
```diff
@@ -0,0 +1,76 @@
...
+       yum install epel-release
+
+#. Install pwauth
+
+    .. code-block:: sh
+
+       yum install pwauth
+
+
+#. Download the mod_authnz_external rpm
```
treydock: This is only needed for RHEL/CentOS 6.

source/authentication/pam.rst
```diff
@@ -0,0 +1,76 @@
...
+
+    .. code-block:: sh
+
+       wget http://download.fedoraproject.org/pub/epel/7/x86_64/Packages/m/mod_authnz_external-3.3.1-7.el7.x86_64.rpm
+
+
+#. Unpack the files from the rpm.
```
treydock: This is only needed for RHEL/CentOS 6.

johrstrom:
OK so given the feedback I should probably add these:

1. A warning about the version # of the RPM saying that EPEL may or may not have this version by the time this is read.
2. Alternative steps to building the .so from source instead of downloading the RPM. Stating that dowloading the RPM is easier but probably less suited for production.

source/authentication/pam.rst
```diff
@@ -0,0 +1,76 @@
...
+
+    .. code-block:: sh
+
+       rpm2cpio mod_authnz_external-3.3.1-7.el7.x86_64.rpm | cpio -idmv
+
+
```
treydock: For RHEL/CentOS 7 there should be mention of `yum install mod_authnz_external` that is done instead of the two commands of `wget` and `rpm2cpio`.

source/authentication/pam.rst
```diff
@@ -0,0 +1,76 @@
...
+      /opt/ood/ood-portal-generator/sbin/update_ood_portal
+
+#. Restart apache httpd
+
+    .. code-block:: sh
+
+      sudo service httpd24-httpd condrestart
+
+#. Cleanup the files extracted in step 4 and disable epel if you wish.
```
treydock: Double check OOD docs, I think we may require EPEL.

treydock:
@johrstrom The best solution is rebuilding the SRPM against SCL Apache for the operating system used at a site but that's a really involved process and would be easier to just build the RPM and provide it from OOD repos than try and document that process.

source/authentication/pam.rst
johrstrom: Users have to dl and unpack because the rpm installation uses / instead of /opt/rh/httpd24/root.  Right? I thought that's why these folks did this.

treydock: Actually kind of.  Part of the issue is they wanted RHEL7 RPM on RHEL6 so only way to do that is download and extract. Also now that I think about it the extract would be necessary on RHEL7 too to avoid installing the system Apache which is almost certainly a dependency. So installing the RPM would be a problem.

source/authentication/pam.rst
treydock: Per other comments, this is needed always to avoid system Apache dependency.

source/authentication/pam.rst
treydock: Per other comments, this is needed always to avoid system Apache dependency.

treydock:
@johrstrom @ericfranz @MorganRodgers 

We may want to evaluate community usage of pre-packaged Apache modules and decide to provide SCL ones as a convenience.  We do this for mod_auth_openidc.

johrstrom:
> provide SCL ones as a convenience
I can't +1 or like that comment, or I would for sure. From what I can discern authentication is a major pain point for installers and anything to make their lives easier I would see as a big win.

johrstrom:
So do we want to keep working? If so, what edits should we make? Do we include how to build the so from the source against the scl httpd? 'Along with' pulling the rpm or 'instead of'?

ericfranz:
For OnDemand 2.0 we can drop support for RHEL6. Which means the system Apache that comes with RHEL7 could meet our needs. However switching to that from SCL Apache will be a non-trivial exercise for most of our users, requiring them to move/copy files from the SCL Apache to the default locations, change or copy and add a new systemd startup file and modify our code to work with Apache configs in the new location. Also, the mapping scripts will need to work with the system Ruby (2.0 ?) instead of the SCL Ruby, since the system Apache won't be configured to load the SCL ruby when executing the mapping scripts (or is that something the Lua code could do, maybe?)

And all of this is up in the air if we replace the lua + mapping scripts + nginx_stage with another solution.

So I say this PR is on hold till we learn more. We can revisit after we make some decisions about the direction of the infrastructure.

johrstrom:
Ok, so you want to keep it open?  I was just going to close - one less thing to keep checking on. Especially if there's no real action items, it'll just become super stale.

ericfranz:
You can close it if you want to capture the ideas here in an issue instead - maybe in the ondemand repo. Something so that we eventually revisit. But yeah, if it stays an open PR I might go back and think "do I need to review this, what is this, oh yeah...".

johrstrom:
> capture the ideas here in an issue instead

I've created [this ticket for supplying mods](https://github.com/OSC/ondemand/issues/46) and a 'discussion' label to facilitate tickets that don't have clear action items or deliverables. I'm a little weary of opening tickets to capture some of the other ideas here just because I would be unable to effectively encapsulate the scope of such tickets and to what end they would be useful, especially if they have very large scopes and very broad impacts.

ericfranz:
The issue you opened and Treys comment and the link to this PR is sufficient for me

ericfranz:
Recently I moved Logout and related authentication documentation up to the Install section to make it more prominent: https://github.com/OSC/ood-documentation/pull/228. In the process I recalled we had an old topic https://discourse.osc.edu/t/implementing-authentication-via-cas/34 where I invited 3 institutions to share their configs for using CAS with OnDemand so I added next to Shibboleth the CAS page with a mention of this Discourse topic as I didn't take time to summarize the different examples. This came up in discussion with @johrstrom  regarding how Discourse has become our documentation and @johrstrom reminded me of this PR, which essentially took the recommendations of a related Discourse topic on PAM and spelled it out here.

It ends up the approach in this PR is similar to the CAS one, with the difference being that the CAS topic suggests to install the module from source using `--prefix=/opt/rh/httpd24/root/etc/httpd/modules`. As I recall the two qualms with this PR were the unpacking of the RPM and the fact that this is how to use PAM with basic auth and we don't want to encourage 2 "bad" practices (using basic auth and unpacking RPM).

For this PR, if we updated it to clone and compile the module with `--prefix=/opt/rh/httpd24/root/etc/httpd/modules` instead of unpacking the RPM, what are your thoughts on adding this to the list? We could add a warning at the top that we do not recommend this approach with using PAM because of the use with BasicAuth and that using another approach (like PAM with Keycloak) would be preferable. But there are certainly situations where the security requirements are so low or are met by a VPN etc. that BasicAuth is an acceptable solution.

Then the 3 modifications we would make would be:

1. Switch to compile from source with `--prefix=/opt/rh/httpd24/root/etc/httpd/modules` 
2. Add warning at top about using BasicAuth and recommend using Keycloak with PAM instead (link to the PAM section of Keycloak documentation)
3. Add a link somewhere on the page to the original Discourse topic

Thoughts?

johrstrom:
Not going to do.

PR #185, opened 2019-08-01
App Sharing

MorganRodgers:
Updates to the app sharing instructions.


johrstrom:
LGTM, though a pre-requisite to sharing an app (through any means) is that the recipient is an ' app developer' right? I would imagine that's worth explicitly stating.

MorganRodgers:
> LGTM, though a pre-requisite to sharing an app (through any means) is that the recipient is an ' app developer' right? I would imagine that's worth explicitly stating.

Good point.

ericfranz:
"File System Permissions" section should be renamed "System Installed Apps" since "Code Sharing" example has now been added

"Examples of App Sharing" should be renamed to "Example of Executable Sharing" and a third subsection of "Peer to Peer Executable Sharing"

source/app-sharing.rst
```diff
@@ -16,61 +18,67 @@ Overview
...
 
-Demo
-----
+    # Given:
+    # - an app named $NEW_APP
+    # - a group named $NEW_APP_GROUP
+    # - and a user named $NEW_APP_USER
 
-This is with two users Eric (efranz) and Bob (an0047).
+    sudo cp "/path/to/$NEW_APP" /var/www/ood/apps/sys
```
ericfranz: Change this to

    sudo cp -r "/path/to/$NEW_APP" "/var/www/ood/apps/sys/$NEW_APP"

source/app-sharing.rst
```diff
@@ -16,61 +18,67 @@ Overview
...
+-------------------------------
 
-Now when Bob accesses the OnDemand home page he sees Eric's MATLAB app and can
-launch it:
-
-.. figure:: /images/app-sharing-6.png
-   :align: center
+By setting a few environment variables it is possible to enable a more polished peer to peer app sharing experience. There are two reasons why this mode is not always enabled: the first is that app permissions are the only thing that prevents all a site's Ondemand users from seeing a shared app, so it is important to get the permissions correct, and only to deploy apps that are production ready. The other reason to be careful with app sharing is that is expands the amount of trust placed in app developers.
 
+.. note:: Executable sharing means the app and all its code runs as the user
```
ericfranz: I would change this back to warning from note.

source/app-sharing.rst
```diff
@@ -16,61 +18,67 @@ Overview
...
+    sudo cp "/path/to/$NEW_APP" /var/www/ood/apps/sys
+    sudo chmod 750 "/var/www/ood/apps/sys/$NEW_APP"
+    sudo chgrp "$NEW_APP_GROUP" "/var/www/ood/apps/sys/$NEW_APP"
 
-Eric has a dev app "Matlab", and interactive plugin app. Eric can
+    sudo usermod -a -G "$NEW_APP_GROUP" "$NEW_APP_USER"
 
-1. Launch Matlab
-2. View and Edit the code
+Using file access lists (FACLs) app sharing may be as specific as desired; sharing with one or more groups, potentially also including or excluding individual users.
```
ericfranz: Read a little awkward. Maybe:

>You can utilize file access lists (FACLs) to increase the granularity with who you share the apps. For example, you could specify multiple groups and individual users to share with, or even exclude specific users or groups.

source/app-sharing.rst
```diff
@@ -16,61 +18,67 @@ Overview
...
 
-.. figure:: /images/app-sharing-1.png
-   :align: center
+Another use for app sharing is sharing an in-development application with members of a development team before deploying that app to the system.
```
ericfranz: I would remove this sentence.

MorganRodgers: Why? It's my primary use for app sharing.

source/app-sharing.rst
```diff
@@ -16,61 +18,67 @@ Overview
    home directory that other users can launch. This is similar to adding to your
    PATH the bin directory of another user.
 
-The rest of this document addresses OnDemand's support for Executable Sharing.
+File System Permissions
+-----------------------
```
ericfranz: At the end of this section, a short paragraph with an example / demo, and then a short paragraph of why this works (changing file permissions on app directories).

For example, you could start by chmod-ing an app to 750 and show the result in the dropdown menu: the app doesn't appear (screenshot 1). Then chgrp of the directory to the group you are in. Reload the dashboard and the link now appears.

The point is that we don't call out here the fact that you can just change the permissions and reload the dashboard in your browser to see the change.

source/app-sharing.rst
ericfranz: This sentence is in a section on System installed apps and app authorization.

ericfranz: If you want to keep it you can reword as follows:

>This app authorization mechanism (using file permissions) can also be useful for canary deployments (sharing an app under development with a subset of users, or just your development team).

ericfranz:
Final thought. We should put a top level page called App Authorization and just explain that 

>Access to apps is controlled with file permissions. See App Sharing documentation for details.

That page will link to the system installed apps documentation or the section on file permissions for controlling access to apps.

screen 2019-08-02 at 11 57 40 AM

PR #187, opened 2019-08-06
[ON HOLD] Rakefile build images

KinanAlAttar:
I have added a docker:build_images task for building the drawio images using drawio-batch. I also added the command '--security-opt label=disable' to the exec line as I was unable to run the rake docker:build command.


Rakefile
```diff
@@ -24,7 +24,12 @@ namespace :docker do
 
   desc "Build docs using docker"
   task :build do
-    exec 'docker run --rm -i -t -v "${PWD}:/doc" -u "$(id -u):$(id -g)" ohiosupercomputer/docker-sphinx make html'
+    exec 'docker run --rm -i -t -v "${PWD}:/doc" -u "$(id -u):$(id -g)" --security-opt label=disable ohiosupercomputer/docker-sphinx make html'
+  end
+  
+  desc "Build drawio images"
+  task :build_images do
+    exec 'docker run --rm -i -t -v "${PWD}:/doc" -u "$(id -u):$(id -g)" --security-opt label=disable ohiosupercomputer/docker-sphinx node ../drawio-batch/drawio-batch.js source/architecture/ood_container_view.drawio source/images/ood_container_view.png && node ../drawio-batch/drawio-batch.js source/architecture/ood_system_view.drawio source/images/ood_system_view.png'
```
MorganRodgers: Thoughts:

- When I tested I didn't need the `--security-opt label=disable` to run, so that may be specific to your set up.
- Instead of running `node ../drawio-batch.js` you can just run `drawio-batch` because it is installed to `/usr/bin/drawio-batch` which is in `$PATH`.
- This should use a `find` command (or similar using Python or NodeJS) that processes found `*.drawio` files instead of processing a hard-coded list.
- I would expect that a single command would both generate the images and then generate the HTML.
- Whatever the solution turns out to be will need to also be implemented for the Makefile since that is what TravisCI is using. That can either be in this PR or a separate PR when we have worked out the issues with this one.

MorganRodgers (changes requested):
I would prefer to leave the Rakefile alone, adding the necessary commands to the Makefile (b/c Travis). At the least this should discover `*.drawio` files. I wouldn't necessarily put all the logic for generating images into the Makefile, I would write a separate Bash, Python or NodeJS file to contain that logic. Note that discovering the files shouldn't be done in Ruby because the container does not have Ruby.

Rakefile
KinanAlAttar: I just comitted and pushed the changes requested.

python/build_images.py
```diff
@@ -0,0 +1,10 @@
+#! /usr/bin/python3.6
+
+import subprocess
+
+string=subprocess.check_output(["find", ".", "-name", "*.drawio", "-print"])
+arr=string.splitlines()
```
MorganRodgers: I assume that `arr` is meant to be array? In Python this is a list, which of course is the name of a builtin function. Better to use more meaningful variables names like `found_files`.

python/build_images.py
```diff
@@ -0,0 +1,10 @@
...
+for s in arr:
+    sD=s.decode("utf-8")
+    subprocess.run(["drawio-batch", sD, 'source/images/'  + sD[22:-7] + ".png"])
```
MorganRodgers: Python 3.6 has [Pathlib](https://docs.python.org/3.6/library/pathlib.html) built in. You should use that for any file path manipulation when working in Python instead of using string indexing, there is a backport for Python2 as well. Then to get the name of a file it becomes just `Path(sD).name` or `Path(sD).stem` if all you want is the file name without its extension.

MorganRodgers: `subprocess.run` needs to have the `check=True` set otherwise it could fail silently.

https://docs.python.org/3.6/library/subprocess.html#subprocess.run

python/build_images.py
```diff
@@ -0,0 +1,10 @@
+#! /usr/bin/python3.6
+
+import subprocess
+
+string=subprocess.check_output(["find", ".", "-name", "*.drawio", "-print"])
+arr=string.splitlines()
+for s in arr:
+    sD=s.decode("utf-8")
```
MorganRodgers: Likewise `sD` is not a useful variable name.

MorganRodgers (changes requested):
`subprocess.run` needs to be fixed so that if it errors we stop the build with an error.

Rakefile
```diff
@@ -21,10 +21,9 @@ namespace :pipenv do
 end
 
 namespace :docker do
-
   desc "Build docs using docker"
   task :build do
-    exec 'docker run --rm -i -t -v "${PWD}:/doc" -u "$(id -u):$(id -g)" ohiosupercomputer/docker-sphinx make html'
+    exec 'docker run --net=host --env="DISPLAY" --volume="$HOME/.Xauthority:/root/.Xauthority:rw" --rm -i -t -v "${PWD}:/doc" -u "$(id -u):$(id -g)" ohiosupercomputer/docker-sphinx make html'
```
MorganRodgers: If you are exporting `DISPLAY` and bind mounting `$HOME/.Xauthority` does this require a running X server on the host?

Makefile
```diff
@@ -43,16 +46,17 @@ help:
 	@echo "  doctest    to run all doctests embedded in the documentation (if enabled)"
 	@echo "  coverage   to run coverage check of the documentation (if enabled)"
 	@echo "  dummy      to check syntax errors of document sources"
+	@echo "  images     to make drawio images"
```
ericfranz: Remove this line, there is no corresponding target to make the images, and I'm not sure we want that at this stage.

Makefile
```diff
@@ -43,16 +46,17 @@ help:
...
 
 .PHONY: clean
 clean:
 	rm -rf $(BUILDDIR)/*
 
 .PHONY: html
 html:
-	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
+	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html && ./$(DRAWIOBUILDPATH)
```
ericfranz: This should be updated to instead just use bash directly instead of python.

```make
html:
         $(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
         find $(BUILDDIR)/html -name "*.drawio" -print | xargs -I {} draw.io -x -f "png" {}
````

File: Rakefile (hunk 5 above)
Comment (KinanAlAttar) [reply Rakefile]: I believe so, yeah.

Comment (MorganRodgers) [reply Rakefile]: Did `drawio-batch` require a running X server as well?

Comment (KinanAlAttar) [reply Rakefile]: I don't think so

Comment (KinanAlAttar) [reply Rakefile]: Just to let you know when I run make html on the container it seems to only work without the user option 
```
 docker run --net=host --env="DISPLAY" --volume="$HOME/.Xauthority:/root/.Xauthority:rw" --rm -i -t -v "${PWD}:/doc" ohiosupercomputer/docker-sphinx make html

```

File: Makefile [hunk 8]
```diff
@@ -51,8 +51,9 @@ clean:
 .PHONY: html
 html:
 	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
+        find $(BUILDDIR)/html -name "*.drawio" -print | xargs -I {} draw.io -x -f "png" {}
```
ericfranz: Remove extra whitespace so it is aligned with the previous statement

Makefile
```diff
@@ -51,8 +51,9 @@ clean:
...
 	@echo
-	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
+	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html & drawio images are in source/images."
```
ericfranz: Revert this message to just: `"Build finished. The HTML pages are in $(BUILDDIR)/html."`

python/build_images.py
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env python3.6
```
ericfranz: This file should be deleted since it is no longer used.

johrstrom:
Not going to do this.

PR #188, opened 2019-08-06
Add documentation for SELinux

treydock:


treydock:
@ericfranz @johrstrom @MorganRodgers Do we need to dig into more detail about the SELinux changes we're making to support SELinux.

treydock:
Also how do we flag a feature as alpha? I don't want people to expect our SELinux support to be 100% functional.

MorganRodgers:
Depending on how broken we think this could be for people I'd use the either the Danger or Warning admonition: http://docutils.sourceforge.net/docs/ref/rst/directives.html#specific-admonitions

ericfranz:
If we merge this into master this rpm should be installable if the repo set is set to https://yum.osc.edu/ondemand/1.6/. If we want to wait until a 1.7 release we can.

MorganRodgers:
Before we ask other sites to use this work I want to try it out against some of our other resource manager VMs.

source/installation/modify-system-security.rst
```diff
@@ -24,6 +27,12 @@ The OnDemand SELinux package makes several changes to allow OnDemand to run with
 * Enable several booleans.
 * Apply a custom policy to allow some additional actions by processes in the `httpd_t` context.
 
+If you experience denails when running SELinux with Open OnDemand please provide denial details by generating a `ood.te` file and posting that to `Discourse <https://discourse.osc.edu/c/open-ondemand>`_. It would also help to post the `audit.log` lines that correspond to the OnDemand specific denials.
```
MorganRodgers: Spelling: 'denials'

treydock: Fixed.

treydock:
The denials in Vagrant will be different and I don't think you can run SELinux in Docker.

I think I can do a 1.6.7 build and just upload the ondemand-selinux package if that's desirable.

MorganRodgers:
Do you have a sense of how the denials will be different in Vagrant? A separate build would be helpful for me.

treydock:
I think main differences would be basic auth as well as maybe difference with home directory being a local /home vs NFS. I don't know if SLURM install location would cause denials. That's the tricky part with supporting SELinux is if a site installs SLURM or Torque to /opt or /var/lib that will change the context of those files and SELinux may block Apache from accessing things.

treydock:
Do we really care about making SELinux work in Vagrant? The main issue I'm facing is around /home/ood being a local home directory and also having weird context because it's a virtualbox mount. I can make policies to allow it but is that worth the effort?

MorganRodgers:
Vagrant is just the environment we have available to us that allows us to test other schedules. The other schedulers are the testing target. Do we have any other way to test other schedulers before opening up to our community without Docker or Vagrant?

treydock:
The only way to reliably test other schedulers would be to install them the exact same way as other sites.  If I do `./configure --prefix=/opt` that will drastically change the SELinux context if I do `./configure --prefix=/usr`. The location of config files and binaries is what matters for SELinux and I don't think we can ensure we test all those possible locations in Vagrant.

I think the best we can do is ensure OnDemand files and policies work and it would be up to a site to make something like SLURM or LSF work with SELinux.

MorganRodgers:
Ok. Then does it make sense for us to ship the Moab-specific rules that we found that we need to other centers? Put another way: do we want to solicit the community for their resource manager specific fixes when they discover them, and eventually ship a works-for-most policy file, or should we only ship a generic policy file and tell people that they should plan to discover what is necessary for their own setup?

treydock:
That's a good point, I've removed the Moab and  Torque specific items from the policy as it would become a support nightmare to expect us to support all schedulers in this way. I think telling people the context of a PUN is `httpd_t` should help them do custom policies specific to their scheduler if they choose to run with SELinux.

MorganRodgers:
I think that probably makes sense.

achalker:
We could probably test them via AWS.


-------------------------
Alan Chalker, Ph.D.
alanc@osc.edu<mailto:alanc@osc.edu>
614-247-8672

source/installation/modify-system-security.rst
```diff
@@ -3,17 +3,40 @@
...
-      certain software in your web applications such as being able to access a
-      Lustre file system.
+      OnDemand runs under the Apache `httpd_t` context.
+
+The OnDemand SELinux package makes several changes to allow OnDemand to run with SELinux enabled.
+
+* Set context of several ondemand-nginx directories and files.
+* Enable several booleans.
+* Apply a custom policy to allow some additional actions by processes in the `httpd_t` context.
+
```
johrstrom: Maybe a link to the source in this block? I feel like if someone enables selinux they'll be interested in knowing what _exactly_ they're changing.

Also I get that a solid link here isn't available yet given we haven't pulled the other PR into master.

ericfranz:
>I think telling people the context of a PUN is httpd_t should help them do custom policies specific to their scheduler if they choose to run with SELinux.

So is the general assumption that if you want to have SELinux enabled, you will know how to craft a policy to enable OnDemand to work with a batch scheduler? A lot of centers installing OnDemand are done by admins who are likely not experts with SELinux.

>That's a good point, I've removed the Moab and Torque specific items from the policy as it would become a support nightmare to expect us to support all schedulers in this way

Maybe instead this work could be sourced by the community. Especially members of the advisory board or other active members who already have test servers who might be more than willing to commit back fixes or additions for each of the schedulers.

Might be an alternative. It is an easier sell to a site installing OnDemand with an SELinux requirement that if they are using Slurm or Moab+Torque someone already has a policy that will help them, versus everyone enabling SELinux must put in the work to craft a custom policy.

treydock:
> So is the general assumption that if you want to have SELinux enabled, you will know how to craft a policy to enable OnDemand to work with a batch scheduler? A lot of centers installing OnDemand are done by admins who are likely not experts with SELinux.

We don't really have much of a choice. If a site installs SLURM into /apps or /opt or /usr/local the contexts between each will be different and some may require custom policies and some may not. If it's a local install vs NFS that will also change things. If a site uses SELinux they have likely already figured out how to make their batch system work with SELinux and will know how to get their install of their batch system to work.

> 
> Maybe instead this work could be sourced by the community. Especially members of the advisory board or other active members who already have test servers who might be more than willing to commit back fixes or additions for each of the schedulers.
> 
> Might be an alternative. It is an easier sell to a site installing OnDemand with an SELinux requirement that if they are using Slurm or Moab+Torque someone already has a policy that will help them, versus everyone enabling SELinux must put in the work to craft a custom policy.

That won't work again because paths may be different. Our Torque policy changes would only work for a site if they install Torque exactly like we do. The Vagrant install of SLURM requires no policy changes because SLURM is in /usr. The issues we faced with Moab and Torque is because of the location of trqauthd socket and the fact we use symlinks for moab.cfg. All of these are very site specific.

I think the best we can do is document how sites can generate their own custom policies, it's really not that hard.  Basic steps:

```
setenforce 0
# Run OnDemand and do things
cat /var/log/audit/audit.log | audit2allow -M my-policy
make -f /usr/share/selinux/devel/Makefile
semodule -i my-policy.pp
```

The likelihood of the OnDemand node being the only node at a site running SELinux is very low. The few people asking for SELinux have it running everywhere so they would already know how to get their batch system to work with SELinux. Technically speaking it's just not practical for us to try and cover every batch system and every possible way those batch systems are installed. If we documented exactly how to install SLURM for OnDemand then it'd make sense for us to also package policies for that install but because a SLURM install is site specific so will the SELinux policies.

Other applications packaging SELinux support are only packaging SELinux support for the things inside their package. It would be very poor practice for us to make policies for things not in the RPM.  It's bad enough we have to enable booleans that assume NFS but we require shared home directories and majority of sites use NFS for that. The alternative is us documenting the booleans site have to enable if they do certain things like NFS home directories.

treydock:
I've added commit to document simple steps for site specific policies.

source/installation/modify-system-security.rst
treydock: Added links to where files will be once things are merged.

PR #191, opened 2019-08-06
Addition of installation from source directions

MorganRodgers:


source/installation/from-source/core-apps.rst
```diff
@@ -0,0 +1,36 @@
...
+  .. _ood-myjobs: https://github.com/OSC/ood-myjobs/
+  .. _ood-shell: https://github.com/OSC/ood-shell/
+
+Each application has its own dependencies which need to be installed (from either NPM or Ruby Gems) by running the following:
+
+  .. code-block:: sh
+
+    cd /var/www/ood/apps/sys/$APP
+    # We have both Node and Rails applications, let's just cover both in a single command
+    sudo NODE_ENV=production RAILS_ENV=production scl enable rh-nodejs6 rh-ruby24 rh-git29 -- bin/setup
```
treydock: If we document one step as using SCL should we document the `yum` command necessary to install those software collections? Maybe in the system-dependencies page?

treydock: Think it would be `yum install rh-nodejs6-npm rh-ruby24-ruby-devel rh-git29-git`. Usually pulling in `-devel` also pulls in runtime.

ericfranz: Yes. I would think that installing from source on a CentOS6 or CentOS7 box would mean using yum install and rpms for the dependencies that are non OnDemand code, and then git cloning the ondemand repos.

Which means, perhaps, the ondemand scl package should be used in the install from source? And if thats the case would the command be `scl enable ondemand -- bin/setup` instead of `scl enable rh-nodejs6 rh-ruby24 rh-git29 -- bin/setup`

MorganRodgers: > ... A use case for some folks wanting this was 'hacking around', I'm not sure how this addresses that, but I'm also not sure _how_ to address that.

My preferred way to do that sort of thing is to spin up a VM using RPMs for OnDemand, then replace the apps that I want to hack on with symlinks to something that is convenient for me to edit locally. The caveat with my workflow with VirtualBox is that the apps do not work if their `vendor/bundle | node_modules` directories are not on the guest. So there is additional symlink-ing necessary and things get messy.

That workflow addresses how people can get a version of OnDemand to hack around with, but isn't satisfying to people who cannot use RPMs.

MorganRodgers: > If we document one step as using SCL should we document the `yum` command necessary to install those software collections? Maybe in the system-dependencies page?

I was trying to avoid becoming too vendor specific. I can flip this the other way and call out specific `yum` package names and versions when necessary.

As for changing the SCL command to use the `ondemand` SCL I can do that, but I'll still call out the specific items that must be in the `$PATH` for non-RPM users.

MorganRodgers:
> ... A use case for some folks wanting this was 'hacking around', I'm not sure how this addresses that, but I'm also not sure _how_ to address that.

My preferred way to do that sort of thing is to spin up a VM using RPMs for OnDemand, then replace the apps that I want to hack on with symlinks to something that is convenient for me to edit locally. The caveat with my workflow with VirtualBox is that the apps do not work if their `vendor/bundle | node_modules` directories are not on the guest. So there is additional symlink-ing necessary and things get messy.

For all its flaws my workflow addresses how people can get a version of OnDemand to hack around with, but isn't satisfying to people who cannot use RPMs.

Reposting here because this was meant to be a response to @johrstrom...

MorganRodgers:
I updated the docs to use OSC provided RPMs for everything except for OnDemand core applications and infrastructure, and confirmed that they work, at least within the context of the `ood-images/vagrant-with-slurm` VM. Given that other parts of the docs cover configuration of OnDemand I think this is done.

Thoughts?

source/installation/from-source/system-dependencies.rst
```diff
@@ -3,42 +3,41 @@
...
+        https://yum.osc.edu/ondemand/1.6/web/el7/x86_64/ondemand-passenger-devel-5.3.7-2.el7.x86_64.rpm \
+        https://yum.osc.edu/ondemand/1.6/web/el7/x86_64/ondemand-passenger-debuginfo-5.3.7-2.el7.x86_64.rpm \
+        https://yum.osc.edu/ondemand/1.6/web/el7/x86_64/ondemand-passenger-5.3.7-2.el7.x86_64.rpm \
+        https://yum.osc.edu/ondemand/1.6/web/el7/x86_64/ondemand-nodejs-1.6-1.el7.x86_64.rpm \
+        https://yum.osc.edu/ondemand/1.6/web/el7/x86_64/ondemand-nginx-filesystem-1.14.0-2.p5.3.7.el7.noarch.rpm \
+        https://yum.osc.edu/ondemand/1.6/web/el7/x86_64/ondemand-nginx-debuginfo-1.14.0-2.p5.3.7.el7.x86_64.rpm \
+        https://yum.osc.edu/ondemand/1.6/web/el7/x86_64/ondemand-nginx-1.14.0-2.p5.3.7.el7.x86_64.rpm \
+        https://yum.osc.edu/ondemand/1.6/web/el7/x86_64/ondemand-git-1.6-1.el7.x86_64.rpm \
+        https://yum.osc.edu/ondemand/1.6/web/el7/x86_64/ondemand-apache-1.6-1.el7.x86_64.rpm \
+        https://yum.osc.edu/ondemand/1.6/web/el7/x86_64/ondemand-1.6.7-1.el7.x86_64.rpm \
```
johrstrom: You're actually installing ondemand here through an rpm. Is that expected?

MorganRodgers: Those are RPMs that OnDemand depends on, but the core RPM for `1.6.x` and `el7` is `ondemand-1.6.7-1.el7.x86_64.rpm`. Maybe I need a better description of what I mean to install from source?

MorganRodgers: ...nvm I see it now...

MorganRodgers: Thanks for catching that Jeff.

PR #194, opened 2019-09-02
Reformat Known Issues

MorganRodgers:
Reformated known issues to be similar to release notes

I removed the note about the file explorer not working well on smaller screens because most of our apps are not fully responsive.

### DIFF
--- a/source/issues/overview.rst
+++ b/source/issues/overview.rst
@@ -9,20 +9,17 @@ OnDemand's Known Issues
 .. File explorer
 
 - `File names with XML or HTML break display`_
-- `Uploading large files is not reliable`_
+- `Problems uploading files`_
 - `Downloading forbidden directories fails silently`_
 - `Viewing large files causes problems`_
-- `File upload may leave an incomplete file`_
 - `Incorrect link in tree root`_
 
 .. Dashboard
 
-- `Rebuild app action in developer view may fail`_
 - `NoMethodError when using a custom translation file`_
 - `Dashboard may be slow due to logic in ERB templates`_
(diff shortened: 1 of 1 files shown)


johrstrom:
I like this all being in one place for sure. I feel like a lot of FAQs or known issues are like that. Are we trying to keep it really short here and link to the github issue for more info? None of these have workarounds listed.

Aside from that - and this isn't necessarily a part of this ticket - none of this stuff shows up in google searches. Having it is one thing, but how are folks supposed to navigate there?  If I search "open ondemand upload fails" none of this stuff shows up.

MorganRodgers:
I don't think that we need to keep this particularly short no. I'll add workarounds for those issues where we have them. Unfortunately in some cases the workarounds are do not use OnDemand which feels bad.

As for discoverability I am not sure that we are big enough for Google to use the Good Algorithms on. The on-page search does okay for the query 'upload fails'.

source/issues/overview.rst
```diff
@@ -1,126 +1,158 @@
...
+- `File upload may leave an incomplete file`_
+- `Incorrect link in tree root`_
 
-		When trying to use certain HTML characters your file name will appear incorrectly. 
-		An issue like this could be resolved by not using the HTML characters. Below are screenshots of what may happen if you do.
+.. Dashboard
 
-		.. figure:: /images/file-names-bad-1.png
-		   :align: center
+- `Rebuild app action in developer view may fail`_
```
ericfranz: Lets just remove the "rebuild app action in developer view". This doesn't affect enough (maybe any) people outside of OSC.

PR #195, opened 2019-09-09
Simplify requirements

treydock:

**Add your description here**

Simplify system requirements. The names for EPEL and SCL were CentOS specific package names so change to just names. The list of packages for OnDemand repo are not requirements, only ondemand package is requirement, rest are either dependencies pulled in by OnDemand package or optional. The list would become out of date very quickly if listing everything in repo.

### DIFF
--- a/source/requirements.rst
+++ b/source/requirements.rst
@@ -8,14 +8,13 @@ Software Requirements
 
 On the Web node serving OnDemand itself:
 
-- `EPEL repository`_
 - `Software Collections repositories`_
 - `lsof`_
 - `sudo`_
 - `OnDemand repository`_:
     - ondemand-{{ondemand_version}}-1.el7.x86_64.rpm
+
 .. _Software Collections repositories: https://www.softwarecollections.org/en/
-.. _EPEL repository: https://fedoraproject.org/wiki/EPEL
 .. _lsof: https://en.wikipedia.org/wiki/Lsof
 .. _OnDemand repository: https://openondemand.org/
 .. _sudo: https://www.sudo.ws/
--- a/source/requirements.rst
+++ b/source/requirements.rst
@@ -8,33 +8,16 @@ Software Requirements
 
 On the Web node serving OnDemand itself:
 
-- `epel-release`_
-- `centos-release-scl`_
+- `EPEL repository`_
+- `Software Collections repositories`_
 - `lsof`_
 - `sudo`_
-- `ondemand-release-web-latest-1-5.noarch.rpm`_:
-    - cjose-0.6.1-1.el7.x86_64.rpm
-    - cjose-debuginfo-0.6.1-1.el7.x86_64.rpm
-    - cjose-devel-0.6.1-1.el7.x86_64.rpm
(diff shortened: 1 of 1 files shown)


treydock:
Can I remove `lsof` and `sudo` since they should be dependencies of OnDemand RPM?  Calling them out when we don't call out every other dependency doesn't seem worth it to me.

johrstrom:
While I agree that it may become over burdensome to state every dependencies version and keep that up to date, I have to wonder what the purpose of such a minimal page would be.  I mean what information is this page really giving to the reader when all it really says is "we depend on ourselves (and epel)"

Maybe we just keep a simpler list some of the larger things that we're going to install and maintain on your (the users) system? Something like this:
* ruby version 2.4 supplied by OSC and installed under SCL
* nginx version 1.14 supplied by OSC and installed under SCL
* etc. 

That way this page serves as an indication of what and how OOD is going to modify their system.

ericfranz:
We could at a link to the corresponding GitHub spec file that identifies these requirements:

https://github.com/OSC/ondemand/blob/v1.6.8/packaging/ondemand.spec#L42-L56

At least it would just be one item to update. It might be easier to write something in the sphinx build steps to ensure this is kept up to date when we increase the version.

ericfranz:
But @treydock you are right when you say:

>The list of packages for OnDemand repo are not requirements, only ondemand package is requirement

I think the primary audience of this page are people installing from rpm, not installing from source, so omitting lsof and sudo and the others is probably appropriate.

source/requirements.rst
```diff
@@ -8,33 +8,17 @@ Software Requirements
...
-    - ondemand-passenger-debuginfo-5.3.7-2.el7.x86_64.rpm
-    - ondemand-passenger-devel-5.3.7-2.el7.x86_64.rpm
-    - ondemand-ruby-{{ondemand_version}}-1.el7.x86_64.rpm
-    - ondemand-runtime-{{ondemand_version}}-1.el7.x86_64.rpm
-    - ondemand-scldevel-{{ondemand_version}}-1.el7.x86_64.rpm
-
-.. _centos-release-scl: https://www.softwarecollections.org/en/
-.. _epel-release: https://fedoraproject.org/wiki/EPEL
+- `OnDemand repository`_:
+    - ondemand-{{ondemand_version}}-1.el7.x86_64.rpm
```
ericfranz: Should this now be `ondemand-{{ondemand_version}}-4.el7.x86_64.rpm`?

treydock: No, that references the actual ondemand RPM. The one using `-4` is `ondemand-release-web`.

ericfranz:
@treydock is this PR still relevant/useful? If so, I think the merge conflict is just due to this change you added after opening this PR:

https://github.com/OSC/ood-documentation/commit/2922dec0ee4251594d51e694bcb6d59cc7fa5559#diff-0c1f55f2f16d7faa34206231e904c447

treydock:
@ericfranz Yes, I rebased and also removed EPEL because we no longer require EPEL.

PR #210, opened 2019-12-02
Document LinuxHost adapter

MorganRodgers:
FWIW: I have a meeting Tuesday, 3rd December to discuss possible changes to the adapter. So this may go out of date quickly.

### DIFF
--- a/source/installation/resource-manager/linuxhost.rst
+++ b/source/installation/resource-manager/linuxhost.rst
@@ -1,0 +1,113 @@
+.. _resource-manager-linuxhost:
+
+LinuxHost Adapter
+=================
+
+.. warning::
+
+   This is an experimental feature right now that may change drastically in the
+   future. So always revisit this documentation after upgrading OnDemand.
+
+First Setup cgroups
+********************
+
+.. warning::
+
+   TODO
+
+How To Configure The Adapter
+****************************
+
+A YAML cluster configuration file for a Linux host looks like:
+
+.. code-block:: yaml
+   :emphasize-lines: 10-
+
+   # /etc/ood/config/clusters.d/my_cluster.yml
+   ---
+   v2:
+     metadata:
+       title: "Owens"
+       url: "https://www.osc.edu/supercomputing/computing/owens"
+       hidden: true
+     login:
+       host: "owens.osc.edu"
+     job:
+       adapter: "linux_host"
+       submit_host: "owens.osc.edu"  # This is the head for a login round robin
+       ssh_hosts: # These are the actual login nodes
+         - owens-login01.hpc.osc.edu
+         - owens-login02.hpc.osc.edu
+         - owens-login03.hpc.osc.edu
+       site_timeout: 7200
+       debug: true
+       singularity_bin: /usr/bin/singularity
+       singularity_bindpath: /etc,/media,/mnt,/opt,/run,/srv,/usr,/var,/users
+       singularity_image: /opt/ood/linuxhost_adapter/centos_7.6.sif
+       # Enabling strict host checking may cause the adapter to fail if the user's known_hosts does not have all the roundrobin hosts
+       strict_host_checking: false  
+       tmux_bin: /usr/bin/tmux
+
+with the following configuration options:
+
+adapter
+  This is set to ``linux_host``.
+submit_host
+  The target execution host for jobs. May be the head for a login round robin. May also be "localhost".
+ssh_hosts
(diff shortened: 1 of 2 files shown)


treydock:
We'll discuss this but can't precreate cgroups when dealing with connections coming from SSH connections that wouldn't also affect regular SSH logins outside LinuxHost adapter.  The solution will be pam_exec like we do on Login nodes at OSC:

In PAM stack

```
session     required      pam_exec.so type=open_session /etc/security/limits.sh
```

/etc/security/limits.sh
```
set -e

PAM_UID=$(id -u "${PAM_USER}")

if [ "${PAM_SERVICE}" = "sshd" -a "${PAM_UID}" -ge 1000 ]; then
        /usr/bin/systemctl set-property "user-${PAM_UID}.slice" \
                MemoryAccounting=true MemoryLimit=64G \
                CPUAccounting=true \
                CPUQuota=700%
fi
```

We'd have to adjust the limits and also add extra check of `PAM_RHOST` to only set these limits when connection comes from OnDemand host.

The cgroup cgrules.conf and cgconfig.conf can only match things by user or group, not by remote host. So the cgroups you set can be for like all users or just admins or just members of a specific group. The rules will assign processes to the necessary cgroup hiearchy. This is what we use in RHEL6 on Ruby but the systemd approach is far superior in terms of catching all processes. The downside to systemd approach is the limits are for all login sessions of a given user so if we say "set limit of 1GB for user foo" then user foo will be limited to 1GB even if they SSH in outside OnDemand and that 1GB limit is shared across all logins on that host.

We MAY be able to use cgrules.conf and set the cgroup based on process name if that's possible with how LinuxHost adapter works. I don't think any other approach I'm aware of will work aside from a custom script we execute via pam_exec that sets custom cgroup values and somehow forces the SSH session into that cgroup.

If the LinuxHost adapter is spawning everything inside Singularity then there are ways with Singularity to set cgroup values.

MorganRodgers:
Thanks Trey, it's good to know what's possible. I added the section about cgroups because the adapter does not place any limits on users, so it would be easy for a "use RStudio as an IDE only" session to turn into a resource intensive session. I would expect that sites would want the cgroups settings for regular SSH sessions and adapter launched processes to be the same. Like Dave says, the point of OnDemand isn't to hide the machine, but I do want a section that makes clear that if sites elect not to setup cgroups then nothing in the adapter will prevent a single user from hogging the machine.

treydock:
I think cgroups are very much a good idea. One more idea I thought of and wanted to write down before I forgot: Wrapping the LinuxHost login node side in a script and setting up cgrules.conf to look for that process. So ideally all interactions would go through like `ondemand-linuxhost-adapter` script and then we could do something like:

```
*:ondemand-linuxhost-adapter  cpu linuxhostadapter
*:ondemand-linuxhost-adapter  memory linuxhostadapter
```

Then in `/etc/cgconfig.d/linuxhostadapter.conf` we'd set something like:

```
group linuxhostadapter {
  memory {
    memory_limit_in_bytes = "1G";
  }
  cpu {
   ...
  }
}
```

The main downside is the cgroup limits would be static and not dynamic, so a site would have to say all linuxhost adapter executions have same limits.

source/installation/resource-manager/linuxhost.rst
```diff
@@ -0,0 +1,113 @@
...
+         - owens-login02.hpc.osc.edu
+         - owens-login03.hpc.osc.edu
+       site_timeout: 7200
+       debug: true
+       singularity_bin: /usr/bin/singularity
+       singularity_bindpath: /etc,/media,/mnt,/opt,/run,/srv,/usr,/var,/users
+       singularity_image: /opt/ood/linuxhost_adapter/centos_7.6.sif
+       # Enabling strict host checking may cause the adapter to fail if the user's known_hosts does not have all the roundrobin hosts
+       strict_host_checking: false  
+       tmux_bin: /usr/bin/tmux
```
MorganRodgers: In the cluster config `tmux_bin` or `singularity_bin` could be used to provide a wrapper which cgroups rules could then target.

treydock:
I was wrong about what `cgconfig.conf` and `cgrules.conf` do. They are limits applied to everyone so on Ruby all users share 108GB which is just a way to avoid crashing the node. I am working on docs for cgroups.

MorganRodgers:
Thanks.

treydock:
Added 2 approaches. One is per-user and one is per-node for all users using this new adapter.

PR #214, opened 2019-12-06
Add Matlab tutorial

### LINKED ISSUE #206: Add directions on setting up an interactive VNC app with a window manager
A separate top level section?

Give two examples, one with MATE one with XFCE - since those are the two desktops we ship support for. The XFCE example is in use at OSC right now https://github.com/OSC/bc_osc_matlab/blob/db4b351d22d47ddf9ecb83a859f044c29b1bf68c/template/script.sh.erb#L11-L26 and https://github.com/OSC/bc_osc_matlab/tree/db4b351d22d47ddf9ecb83a859f044c29b1bf68c/template/config

For MATE, there is marco, but then there are other tools that work with it. For example, you need a panel of some type so when you minimize a window managed by marco you can find the minimized window and maximize it again.

What are the steps for this?

MorganRodgers:
Fixes #206.

### DIFF
--- a/source/app-development/tutorials-interactive-apps/add-matlab/deploy.rst
+++ b/source/app-development/tutorials-interactive-apps/add-matlab/deploy.rst
@@ -1,0 +1,29 @@
+.. _app-development-tutorials-interactive-apps-add-matlab-deploy:
+
+Deploy RStudio App
+==================
+
+After you have confirmed you are happy with all of your app changes you can
+deploy it as a System App.
+
+.. warning::
+
+   The following steps **MUST** be done from the OnDemand host and not another
+   machine.
+
+#. Go to your OnDemand sandbox directory:
+
+   .. code-block:: sh
+
+      cd ~/ondemand/dev
+
+#. Copy the app to the system deployment location as **root**:
+
+   .. code-block:: sh
+
+      sudo cp -r bc_my_center_matlab /var/www/ood/apps/sys/.
+
+#. In your browser navigate to the OnDemand URL and confirm you see the new app
+   in the *Interactive Apps* dropdown on the :ref:`dashboard`::
+
+     https://ondemand.my_center.edu/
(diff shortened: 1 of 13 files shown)


source/app-development/tutorials-interactive-apps/add-matlab/deploy.rst
```diff
@@ -0,0 +1,29 @@
+.. _app-development-tutorials-interactive-apps-add-matlab-deploy:
+
+Deploy RStudio App
```
johrstrom: Says RStudio here.

johrstrom:
I think this seems fine in and of itself, but there's a lot of duplicate information in here. After this is pulled in, should we create a ticket(s) to refactor this a little bit? Like the `form.js` is very good, but when I searched our docs for it, (a) I couldn't find it, and (b) would be very helpful as a standalone entry. Same for the desktop info.

ericfranz:
@johrstrom what are you thinking of in particular that is "duplicate"?

I think that the idea of having a stand-alone on form.js and different windowing environments as separate sections in the documentation on building the plugins may be useful and to link to those from here.

Also I wonder, Mathwork's original question was: when an OnDemand site has users who experience problems such as the java errors or the window manager issues, what link should we provide them? Is this tutorial that link, or do we need to carve out the specific information relevant for people facing those issues to quickly answer those questions?

johrstrom:
Duplicate was probably the wrong word. Some good information just applies to interactive apps in general not matlab specifically, and we should move them around after this PR is closed.

To the other question, I'd say maybe a troubleshooting section within this tutorial with known issues/problems would be helpful.

MorganRodgers:
There's a fair amount of truly duplicate content copied from the Jupyter tutorial. I think that the question of whether our community is better served by beginning to end tutorials or tutorials that focus on the new stuff.

I'm inclined to believe that the duplication is acceptable because I do not expect that admins will be interested in installing all the apps.

MorganRodgers:
@johrstrom by a troubleshooting / known issues section would moving the if-you-get-java-errors section be sufficient or did you have something else in mind?

johrstrom:
Yea that was my thinking, that bit being in it's own section/page instead of under `edit-script-sh`. 

That way it's a part of this tutorial (as a reminder to the person who installed it that they can look back here when they get into trouble) but also it's own link to provide to Mathworks (or even folks on discourse).

PR #215, opened 2019-12-06
add docs on enabling google analytics

johrstrom:
Add docs for enabling google analytics.  

Not sure if it should be a top level item in the index, but also didn't really know where else to put it.


treydock:
Maybe as a section in customization, if not at top level?

ericfranz:
Remember https://osc.github.io/ood-documentation-test/add-google-analytics/google-analytics.html

Instead of a top level `Install > Adding Google Analytics` I would do `Install > Analytics > Adding Google Analytics` so `Install > Analytics` is the top section.

That way I can add a page to the `Install > Analytics` section that is basically the copy of what I had sent in an email to some sites about alternatives to Google Analytics i.e. `Parsing Apache Logs` and `Parsing Per User NGINX Logs` and then maybe any other relevant or helpful information that @treydock wants to add about syslog and Splunk etc.

source/analytics/google-analytics.rst
```diff
@@ -0,0 +1,167 @@
...
+
+If you wish you can setup your Open-OnDemand instance to send usage data to Google Analytics
+(GA) that you can then query and report on, this page walks through how to do just that.
+
+.. note::
+
+  You'll need to have a `Google Analytics`_ account setup as a prerequisite to this.
+
+  To query GA You'll need to have to have a service account setup with the appropriate permissions. See
+  info on `Google service accounts`_ and `Google iam roles`_ for more details on that.
```
MorganRodgers: IAM is an acronym and should be capitalized.

source/analytics/google-analytics.rst
```diff
@@ -0,0 +1,167 @@
...
+
+
+Configure Open-OnDemand
```
MorganRodgers: Open OnDemand doesn't have a hyphen.

source/analytics/google-analytics.rst
```diff
@@ -0,0 +1,167 @@
...
+  end
+
+
+More Info
+-----------
+
+.. _GA measurement protocol: https://developers.google.com/analytics/devguides/collection/protocol/v1/reference
+.. _analytics lua code: https://github.com/OSC/ondemand/blob/master/mod_ood_proxy/lib/analytics.lua
+
+See our `analytics lua code`_ for the implementation of how we're extracting this information,
```
MorganRodgers: I'm glad you included this; to clarify this is included only for reference, and not because the user is expected to want/need to change anything?

johrstrom: It was for reference.  I don't _expect_ folks to need to change anything, but who knows?

MorganRodgers: You might just add the 'For reference' to the start of line 164?

PR #223, opened 2019-12-20
Add troubleshooting page for interactive apps

### LINKED ISSUE #219: Add FAQ for common issues when setting up interactive apps
In particular: requirement/reason for window management for Lisa from Mathworks

Other ideas include:

- Tips and tricks with `form.js`
- Waiting for a backgrounding job
- Changing defaults
- Safe and unsafe uses for ERB in the `$PLUGIN/template` directory

MorganRodgers:
https://osc.github.io/ood-documentation-test/troubleshooting-interactive-apps/tutorials-interactive-apps/troubleshooting.html

Starts to fix #219.


MorganRodgers:
Maybe this should be under known issues?

ericfranz:
I think of “known issues” as bugs that currently are not or cannot be fixed. So I think this should be in a separate section like you have it

PR #227, opened 2020-01-02
Merge matlab window mrg examples

ericfranz:
https://osc.github.io/ood-documentation-test/merge_matlab_window_mgr_examples/app-development/tutorials-interactive-apps/add-matlab/edit-script-sh.html

This merges the three window mgr examples into the page on editing the script file that starts MATLAB. 

Benefits:
- This improves readability of the window manager changes by seeing their code in context with the rest of the code.

Drawbacks:
- This results in duplication of the MATLAB block on this page - there are now 4 duplicated sections of it.


MorganRodgers:
I think that this is fine. The duplication is annoying but not the end of the world and it reduces the chance that someone is going to copy and paste something incorrectly.

PR #245, opened 2020-03-11
add a google analytics tracker to the layout

johrstrom:

https://osc.github.io/ood-documentation-test/add-ga-tracker

This adds a google analytics tracker so we can see how folks navigate our docs.   Note that I want to pull this into master and not develop. After this, we probably want to pull it into some, if not all, release- branches and develop maybe too.

### DIFF
--- a/source/_templates/layout.html
+++ b/source/_templates/layout.html
@@ -1,0 +1,13 @@
+{% extends "!layout.html" %}
+
+{% block extrahead %}
+{{ super() }}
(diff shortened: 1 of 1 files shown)


ericfranz:
So instead of going through the exercise of reverting the maintenance mode docs merge on develop, adding to 1.7 and then ensuring that develop and master are the same, I'm going to instead do the following: 

1. merge this into master as you suggested
2. cherry-picking this commit to develop, release-1.7, old release branches
3. update develop so it matches release-1.7

what this means then is we will not be updating the master branch of the documentation with any new fixes until we announce a stable release 1.7.

PR #247, opened 2020-03-20
update LHA documentation

johrstrom:
https://osc.github.io/ood-documentation-test/update-lha-w-desktop

* add a troubleshooting entry in the linuxhost.rst file
* add a little bit of docs around enabling a LHA backed desktop

Really these are just some linux host adapter updates for troubleshooting and creating a bc desktop with it.


source/enable-desktops/modify-form-attributes.rst
```diff
@@ -280,5 +280,36 @@ value of ``8`` in the "Number of hours" form field.
...
+   cluster: "owens_login"
+   attributes:
+   desktop: "xfce"
+   bc_num_hours:
+      value: 1
+   bc_num_slots: null
+   bc_queue: null
+   bc_account: null
+   node_type: null
+   bc_email_on_started: null
```
ericfranz: the attributes need indented underneath `attributes:` section i.e.

```
attributes:
  bc_num_hours:
    value: 1
  bc_num_slots: null
```

I wonder if a replacement of the `form:` section itself is preferable to specifying null for all the unused fields. I think if you do this you get the same outcome:

```yaml
---
title: Owens Login XFCE desktop
description: This launches a XFCE desktop on an Owens login nodes.
cluster: owens_login
form:
  - desktop
  - bc_num_hours
attributes:
  bc_num_hours:
    value: 1
  desktop: "xfce"
```

I do not know which is preferable to recommend.

johrstrom: Yea I think specifying the form as you have it is a lot cleaner. Updated for the same.

PR #248, opened 2020-03-23
add descriptive docs for maintanence mode

johrstrom:
https://osc.github.io/ood-documentation-test/add-maintanence-info

Add some more info around maintenance mode.

### DIFF
--- a/source/customization.rst
+++ b/source/customization.rst
@@ -630,3 +630,34 @@ If the balance is defined as a ``project`` balance, then it applies to a project
      "project": "project1",
      "value": 10
    }
+
+Maintenance Mode
+-----------------
+
+As an administrator you may want to have some downtime of the Open OnDemand service for various reasons,
+while still telling your customers that the downtime is expected.
+
+You can do this by setting Open OnDemand in 'Maintenance Mode'. Apache will serve 
+``/var/www/ood/public/maintenance/index.html`` which you can change or brand to be your own. Changes
+to this file will persist through upgrades.
+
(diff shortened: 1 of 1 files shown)


source/customization.rst
```diff
@@ -630,3 +630,34 @@ If the balance is defined as a ``project`` balance, then it applies to a project
...
+-----------------
+
+As an administrator you may want to have some downtime of the Open OnDemand service for various reasons,
+while still telling your customers that the downtime is expected.
+
+You can do this by setting Open OnDemand in 'Maintenance Mode'. Apache will serve 
+``/var/www/ood/public/maintenance/index.html`` which you can change or brand to be your own. Changes
+to this file will persist through upgrades.
+
+Apache return this html file and a 503 response code to all users who's ip does not match one of the
```
treydock: `s/Apache return/Apache returns/g` , plural returns.  Also probably good to capitalize `ip` to be `IP`.

treydock:
Administrators should NOT modify `ood-portal.yml` to enable/disable maintenance mode as that will produce unnecessary restarts of Apache.  They need to touch `/etc/ood/maintenance.enable` and when maintenance is over they remove `/etc/ood/maintenance.enable`.  The Apache config should only be modified to add/remove whitelist addresses or to completely disable the ability to do maintenance modes via touching a file.

So I'd recommend removing part about changing Apache configs around maintenance enable/disable and making a section or comments about enabling/disabling maintenance using the file `/etc/ood/maintenance.enable`.

treydock:
Looks good.

PR #255, opened 2020-04-17
documentation for GE's use of OOD_JOB_NAME_ILLEGAL_CHARS

johrstrom:
https://osc.github.io/ood-documentation-test/ge-job-names

Docs for #246.  The related tickets mention PBSPro as well, but I tracked that down to [this ticket](https://github.com/OSC/ondemand/issues/341) in which the issue was due to colons in the output path (-o option).  As it stands, this feature only [sanitizes job names](https://github.com/OSC/ondemand/pull/429/files#diff-9ddf794da41002473c7abf3dc34e48d5R242). 

Which is to say, I've only documented this feature as it relates to Grid engine and left PBSPro alone.

### DIFF
--- a/source/installation/resource-manager/sge.rst
+++ b/source/installation/resource-manager/sge.rst
@@ -101,9 +101,12 @@ job names like the error below.
 
 ``Unable to read script file because of error: ERROR! argument to -N option must not contain /``
 
-To resolve this set ``OOD_JOB_NAME_ILLEGAL_CHARS`` in the ``/etc/ood/config/apps/dashboard/env`` file.
+To resolve this set ``OOD_JOB_NAME_ILLEGAL_CHARS`` to ``/`` for all OOD applications in the 
+``/etc/ood/profile`` file.
 
 .. code-block:: text
 
-   # /etc/ood/config/apps/dashboard/env
+   # /etc/ood/profile
    OOD_JOB_NAME_ILLEGAL_CHARS="/"
+
+   . /opt/ood/nginx_stage/etc/profile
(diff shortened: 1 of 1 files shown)


source/installation/resource-manager/sge.rst
```diff
@@ -93,3 +93,20 @@ This sets the sh shell to behave like bash and ensures you've sourced your users
...
+
+Invalid Job name
+****************
+You're likely to encounter an issue in running batch connect applications complaining about invalid
+job names like the error below.
+
+``Unable to read script file because of error: ERROR! argument to -N option must not contain /``
+
+To resolve this set ``OOD_JOB_NAME_ILLEGAL_CHARS`` to ``/`` for all OOD applications in the 
+``/etc/ood/profile`` file.
```
ericfranz, line 105: `/etc/ood/profile` is not the correct place for these type of modifications. Rather, `/etc/ood/config/nginx_stage.yml` and `pun_custom_env`

We might need to add clarification to the documentation in https://osc.github.io/ood-documentation/master/customization_overview.html to move `/etc/ood/profile` to clarify the first items are for changing the environment of the per user NGINX for various OnDemand configuration options, and that /etc/ood/profile is special for changing things like SCLs to load, etc. I can do that.

source/installation/resource-manager/sge.rst
```diff
@@ -93,3 +93,20 @@ This sets the sh shell to behave like bash and ensures you've sourced your users
             set +o posix
             . ~/.bashrc
             %s
+
+Invalid Job name
+****************
+You're likely to encounter an issue in running batch connect applications complaining about invalid
```
ericfranz, line 99: >You're likely to encounter an issue in...

change to

>If you encounter an issue with invalid job names when running batch connect like the error below:

The truth is if you are using SGE you are not likely to experience this. Many sites have been running OnDemand with SGE for a while now and we only recently heard of this issue with illegal job name, so it only affects a small subset of the SGE sites. That may be due to the SGE flavor UGE being more forgiving.

PR #257, opened 2020-04-17
add cors documentation for the shell app

johrstrom:
https://osc.github.io/ood-documentation-test/add-cors-1.7

OOD_SHELL_ORIGIN_CHECK docs for release 1.7.

### DIFF
--- a/source/customization.rst
+++ b/source/customization.rst
@@ -245,7 +245,7 @@ In ``/etc/ood/config/apps/shell/env`` set the env var ``DEFAULT_SSHHOST`` to cha
 This will control what host the shell app ssh's to when the URL accessed is ``/pun/sys/shell/ssh/default`` which is the URL other apps will use (unless there is context to specify the cluster to ssh to).
 
 Set Shell App Origin
---------------
+--------------------
 
 The OOD shell application verifies the origin of a request to protect against malicious CORS_ attacks.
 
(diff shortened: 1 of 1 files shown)


source/customization.rst
```diff
@@ -244,6 +244,24 @@ In ``/etc/ood/config/apps/shell/env`` set the env var ``DEFAULT_SSHHOST`` to cha
 
 This will control what host the shell app ssh's to when the URL accessed is ``/pun/sys/shell/ssh/default`` which is the URL other apps will use (unless there is context to specify the cluster to ssh to).
 
+Set Shell App Origin
+--------------------
+
+The OOD shell application verifies the origin of a request to protect against malicious CORS_ attacks.
```
ericfranz, line 250: Re title the section to: "Fix Unauthorized WebSocket Connection in Shell App" and first sentence:

>If you see a 401 error when attempting to launch a Shell app session, where the request URL starts with `wss://` and the response header includes `X-OOD-Failure-Reason: invalid origin`, you may need to set the `OOD_SHELL_ORIGIN_CHECK` configuration option.

Then the section use the bulk of the text I wrote on [the Discourse announcement](https://discourse.osc.edu/t/security-fix-in-open-ondemand-patch-release-1-6-22-and-1-5-7-now-available/819) including details about X-Forwarded-Proto and X-Forwarded-Host and the use of CSRF token check.

Right now this doesn't mention the CSRF token check at all so the implication is that the only protection from CSRF is the origin check and it might scare people to disable it.

Also, it should be called CSRF vulnerability, not CORS attack, and the link to https://owasp.org/www-community/attacks/csrf to be used.

PR #263, opened 2020-04-21
Clarify purpose of /etc/ood/profile

ericfranz:
https://osc.github.io/ood-documentation-test/clarify_etc_ood_profile/customization_overview.html

Group into two sections, the ones intended for changing PUN and app environment, and the one intended for swapping out SCLs.

### DIFF
--- a/source/customization_overview.rst
+++ b/source/customization_overview.rst
@@ -23,7 +23,7 @@ These are three files that directly affect the environment or functionality of t
      - replace ``$APP`` above with the name of the app directory deployed to ``/var/www/ood/apps/sys``
      - Modify Rails application behavior `using Ruby code <https://guides.rubyonrails.org/configuring.html#using-initializer-files>`__. Since this is application code environment variables can be set, removed.
      - This method is specific to Ruby on Rails applications: Activejobs,
-       Dashboard, File Editor, and Job Composer. You can add multiple
+       Dashboard, and Job Composer. You can add multiple
        initializer files in this directory and they will be loaded in
        alphabetical order
 
--- a/source/customization_overview.rst
+++ b/source/customization_overview.rst
@@ -5,10 +5,7 @@ Customization Overview
 
 OnDemand's configuration is stored in the environment, which can be modified using a hierarchy of config files. Configuration files are rooted in ``/etc/ood``. Assets that should be publicly available (such as a favicon or logo image) are placed under ``/var/www/ood/public``.
 
-   * **/etc/ood/profile**
-
-     - If exists, this file is sourced instead of the default at ``/opt/ood/nginx_stage/etc/profile`` by ``/opt/ood/nginx_stage/sbin/nginx_stage`` script when running as root, prior to launching the PUN.
-     - You should source ``/opt/ood/nginx_stage/etc/profile`` in your custom ``/etc/ood/profile`` if you add one to load the correct software collections.
+These are three files that directly affect the environment or functionality of the Per-User NGINX (PUN) processes:
 
    * **/etc/ood/config/nginx_stage.yml**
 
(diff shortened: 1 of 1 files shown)


source/customization_overview.rst
```diff
@@ -17,13 +14,35 @@ OnDemand's configuration is stored in the environment, which can be modified usi
...
+     - replace ``$APP`` above with the name of the app directory deployed to ``/var/www/ood/apps/sys``
      - Used to provide application specific config
      - ``env`` files do not override values set by prior methods.
 
    * **/etc/ood/config/apps/$APP/initializers/ood.rb**
 
+     - replace ``$APP`` above with the name of the app directory deployed to ``/var/www/ood/apps/sys``
      - Modify Rails application behavior `using Ruby code <https://guides.rubyonrails.org/configuring.html#using-initializer-files>`__. Since this is application code environment variables can be set, removed.
      - This method is specific to Ruby on Rails applications: Activejobs,
-       Dashboard, File Editor, and Job Composer. You can add multiple
```
johrstrom, line 27: The file editor is a rails app. Did you mistake this for the file browser?

ericfranz: This is intentional. I don't want to encourage people to add initializers to the file editor. I doubt anyone has. In 1.8 I intend to finally pull in Tyler's changes to make this a Sinatra (or simpler) app since Rails is unnecessary.

johrstrom:
It's a step in the right direction for sure because it's not immediately apparent. I guess the confusion is around profile and that being the usual place to set stuff for shell related environments, but in our case we want to stress the use of nginx_stage.yml.

Would recipes help here? Like
* Global static environment variable:  declare in nginx_stage.yml
```yaml
# /etc/ood/config/nginx_stage.yml
pun_custom_env:
    FOO: "BAR"
```
* App specific environment variable: declare in /etc/ood/config/apps/$APP/env
```text
# /etc/ood/config/apps/$APP/env
FOO=BAR
```
* Global dynamic environment variable: create in profile and declare in nginx_stage.yml
...

Also why the on focus SCLs? Isn't it for _anything_ that's dynamically generated like SCLs?

ericfranz:
The primary purpose of the `/etc/ood/profile` was to enable us to change the SCLs loaded prior to starting NGINX.

The file is sourced by root. So it is not appropriate for anything dynamically generated that should be done as an unprivileged user.

I'll add the two recipes as suggested.

johrstrom:
For Cloudy cluster I write out this `/etc/ood/profile`, I guess I should say, 'evaluated' instead of 'dynamic'.  This was the third recipe just noted as ellipses for brevity. And yes, the remote_user is not known (I think we've tried resetting $HOME here?). 

Is this the appropriate place for this?
```shell
#!/bin/bash
PATH=$PATH:/opt/orangefs/bin:/opt/orangefs/sbin:/opt/CloudyCluster/srv/:/opt/CloudyCluster/srv/CCQ
source /opt/ood/nginx_stage/etc/profile
```

ericfranz:
If binaries in all of these paths are required for Cloudy Cluster, and we follow the convention of the other job adapters, the way to do that would be to add this to the cluster config. For example, the Slurm config includes like:

```yaml
job:
  adapter: "slurm"
  cluster: "my_cluster"
  bin: "/path/to/slurm/bin"
  conf: "/path/to/slurm.conf"
```

and cloudy cluster could have:

```yaml
job:
  adapter: "cloudycluster"
  srv: "/opt/CloudyCluster/srv"
  ccq: "/opt/CloudyCluster/srv/CCQ"
  orangefs_bin: "/opt/orangefs/bin"
  orangefs_sbin: "/opt/orangefs/sbin"
```

Then the adapter would use these to build the full path to the command it executes.

PR #269, opened 2020-04-22
Document rpm size reduction & ondemand-gems

ericfranz:
https://osc.github.io/ood-documentation-test/ondemand_gems_rpm_in_release_notes/

### DIFF
--- a/source/release-notes/v1.7-release-notes.rst
+++ b/source/release-notes/v1.7-release-notes.rst
@@ -22,6 +22,8 @@ Highlights in 1.7:
 - `Dashboard now alerts on malformed cluster configs`_
 - `Job composer now shows suggested scripts`_
 - `Add ability to disable SSH to compute node link in batch connect session card`_
+- `Reduced rpm size by 40 percent`_
+- `New shared OnDemand gem sets`_
 
 Special thanks
 --------------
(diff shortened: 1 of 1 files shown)


ericfranz:
We had a significant reduction in the rpm size (40%) and thus the resulting space taken up by an OnDemand install. I haven't determined the latter. A quick way to do that to include here? Otherwise is this okay?

If so I'll merge. I'll announce 1.7 tonight if you have no other objections to the documentation or the release announcement.

johrstrom:
Maybe this can be in a different ticket, but now `$GEM_HOME` is important now because of this. That's what tripped up these folks who use a cgexec wrapper. 

https://discourse.osc.edu/t/failed-upgrade-1-6-19-1-7-6-cannot-load-such-file-bundler-setup-loaderror/788/4

source/release-notes/v1.7-release-notes.rst
```diff
@@ -289,4 +291,18 @@ Administrators can now disable the link that appears in the batch connects' card
...
+
+The 1.6 ondemand-1.6.22-1.el7.x86_64.rpm was 162M and in 1.7 this has been reduced by 63M to a total of 99M, split between two rpms: ondemand-gems-1.7.10-1.7.10-2.el7.x86_64.rpm with 68M, and ondemand-1.7.10-2.el7.x86_64.rpm with 31M. This is in part due to switching to a monorepo at https://github.com/osc/ondemand and installing all the gems into a central shared location, installed using the new ondemand-gems rpm.
+
+New shared OnDemand gem sets
+............................
+
+The new ondemand-gems rpms allow other Passenger apps to be developed using the same gems that the core OnDemand apps use. If you do ``scl enable ondemand`` and then ``gem list`` you can see all the gems available.
+
+Because the ondemand-gems rpms will include the version in the gem name, it is possible to install multiple versions of these rpms on the same system. For example, "ondemand-gems-1.7.10" is the name and "1.7.10-2.el7.x86_64.rpm" the version of ondemand-gems-1.7.10-1.7.10-2.el7.x86_64.rpm. That way you can upgrade to newer versions of OnDemand but still have the ability to provide apps access to older gems by installing the older gem sets.
```
treydock, line 308: Replace `include the version in the gem name` with `include the version in the RPM name`.

Also `rpms` should be `RPMs`.

treydock:
The size of OnDemand on disk for 1.7.10:

```
[root@web02 ~]# du -h -s  -x /var/www/ood/apps/sys /opt/ood 
290M   /var/www/ood/apps/sys
1.5G   /opt/ood
```

Now some of that is stuff we don't publish since we install many apps and in some cases extra gems with those apps that are not a part of the public version.  RPMs themselves are highly compressed.  That du command could be used on fresh installs to determine size or compare the size of a docker container from OnDemand 1.6 to 1.7 (not yet built).

ericfranz:
@treydock said he'd take this over

treydock:
> https://discourse.osc.edu/t/failed-upgrade-1-6-19-1-7-6-cannot-load-such-file-bundler-setup-loaderror/788/4

There is a lot wrong with how that got implemented, they bypass the ondemand SCL load entirely which is going to break a lot of things.  Not only does that break GEM_HOME it will also break GEM_PATH. I think instead of spelling out every environment variable that's important we need to tell people that they need to load `ondemand` SCL always to avoid issues.  I think that's maybe a different pull request as I'm not sure where that fits in as that's not something new to 1.7.

treydock:
Realized for source installs maybe we do have to spell out environment variables or maybe provide SCL-like load script so we can still utilize our SCL load script.

ericfranz:
For build from source it is reasonable for a site to build assets into `vendor/bundle` of the app repo as before. All the tricks we are playing with GEM_HOME are really specific to our SCLs and our rpms.

PR #271, opened 2020-05-08
Add reference

### LINKED ISSUE #231: Move relevant legacy docs to more prominenant locations in documentation
For example see https://osc.github.io/ood-documentation-test/logout/infrastructure/ood-portal-generator.html

Some of these we already address in the documentation, such as the Example Configurations section, though reviewing each to verify all the important information is in the other spots should be done.

johrstrom:
https://osc.github.io/ood-documentation-test/add-reference/

Add a reference section. Attempt to resolve #231 but also could be a place to add additional reference documentation for `form.yml` and `submit.yml.erb`.  

The submit I think is important because there are a lot of options in `script` or `batch_connect` that folks simply don't know they can override and having a plain old reference section detailing this may be helpful.


johrstrom:
Do we have thoughts on this?  I copied the `ood-portal-generator` directly and it's `bin/generate` instead of `sbin/update_ood_portal` so I know that could change.

source/reference/files/ood-portal-yml.rst
```diff
@@ -1,7 +1,7 @@
 .. _ood-portal-generator-configuration:
 
-Configuration
-=============
+ood-portal.yml
```
ericfranz, line 3: ood_portal.yml should be the title, not ood-portal.yml since the file is ood_portal.yml

ericfranz:
In https://osc.github.io/ood-documentation-test/add-reference/reference/files/ood-portal-yml.html the first line suggest using an example file as a starting point:

    sudo cp share/ood_portal_example.yml /etc/ood/config/ood_portal.yml

This should be updated to reflect the new monorepo and install location of these files: /opt/ood/ood-portal-generator/ood_portal_example.yml

Also I wonder if this is actually auto generated on a new install now. If that is the case, a copy is no longer required and instead it should be stated that a file was dropped here to be edited, and an example file can be referenced at /opt/ood/ood-portal-generator/ood_portal_example.yml

ericfranz:
On the ood-portal-generator page https://osc.github.io/ood-documentation-test/add-reference/reference/commands/ood-portal-generator.html this is problematic:


>The command that generates the Apache configuration file is given as:
>
>     bin/generate [OPTIONS]

Rather it is

    /opt/ood/ood-portal-generator/bin/generate

But even though that command exists, it is done only the first time, and then every time after that it is as you state above update_ood_portal. In fact, if you install via rpm, you would probably only ever do update_ood_portal.

It is also important to note that we changed the way Apache starts so that update_ood_portal is executed prior to starting Apache - so the config is updated every time you restart Apache.

So this is an example of something missing from these references - at the top there should be a section, maybe right after the summary and before the details of the options, that is "When would I edit this file?" or "When would I run this program?"

ericfranz:
Project and code and library documentation may be useful but I think that Reference here is higher level than that. Reference here seems to be targeted towards providing helpful information to admins that is actionable. The mod_ood_proxy section should be removed from this, because it is documentation about an implementation detail of OnDemand, but itself is not actionable. Utilizing or customizing the user of those handlers is done in ood_portal.yml.

ericfranz:
For the ood_auth_map directory with the two scripts, there is so little information there that it could be all moved to the Authentication section. So we could remove the ood_auth_map section from Reference as well.

ericfranz:
Another thought on this Reference section - a lot of this detailed documentation can become easily out of sync with the code itself - it will be harder to keep this up to date. It would be worth looking into options for auto-generating reference documentation from a specified set of example configuration files, documentation of the scripts themselves, etc.

johrstrom:
I've updated from the comments.  I've replaced `bin/generate` with `sbin/update_ood_portal`.  I don't believe users need to use `bin/generate` at all. Removed mod-ood-proxy and moved ood-auth-map section to `authentication/setup auth map` because that section already existed.

johrstrom:
Can this be included in 1.8?

PR #314, opened 2020-07-21
update customization information for shell app

matthu017:
OOD_DEFAULT_SSHHOST and OOD_SSHHOST_ALLOWLIST specified in env and how they work

Also details changes from future PRs https://github.com/OSC/ondemand/pull/601 and https://github.com/OSC/ondemand/pull/570

### DIFF
--- a/source/customization.rst
+++ b/source/customization.rst
@@ -252,9 +252,16 @@ We recommend setting this environment variable in ``/etc/ood/config/nginx_stage.
 Set Default SSH Host
 --------------------
 
-In ``/etc/ood/config/apps/shell/env`` set the env var ``DEFAULT_SSHHOST`` to change the default ssh host. Otherwise it will default to "localhost" i.e. add the line ``DEFAULT_SSHHOST="localhost"``.
+In ``/etc/ood/config/apps/shell/env`` set the env var ``OOD_DEFAULT_SSHHOST`` to change the default ssh host. Otherwise it will default to the first cluster yaml file with ``default: true`` under login. Else it will default to the first cluster yaml file that does not have ``hidden: true``
 
-This will control what host the shell app ssh's to when the URL accessed is ``/pun/sys/shell/ssh/default`` which is the URL other apps will use (unless there is context to specify the cluster to ssh to).
+This will control what host the shell app ssh's to when the URL accessed is ``/pun/sys/shell/`` to ``/pun/sys/shell/ssh/default`` which is the URL other apps will use (unless there is context to specify the cluster to ssh to).
+
+Set SSH Host Allowlist
+----------------------
+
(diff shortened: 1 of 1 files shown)


source/customization.rst
```diff
@@ -252,9 +252,16 @@ We recommend setting this environment variable in ``/etc/ood/config/nginx_stage.
 Set Default SSH Host
 --------------------
 
-In ``/etc/ood/config/apps/shell/env`` set the env var ``DEFAULT_SSHHOST`` to change the default ssh host. Otherwise it will default to "localhost" i.e. add the line ``DEFAULT_SSHHOST="localhost"``.
+In ``/etc/ood/config/apps/shell/env`` set the env var ``OOD_DEFAULT_SSHHOST`` to change the default ssh host. Otherwise it will default to the first cluster yaml file with ``default: true`` under login. Else it will default to the first cluster yaml file that does not have ``hidden: true``
```
johrstrom, line 255: I'll have to dig into this more, but I think that there is no base out of the box default anymore.

If so, we should add that it no longer will default to `localhost`. You have to set it somewhere through the environment variables or through cluster yamls.

source/customization.rst
```diff
@@ -252,9 +252,16 @@ We recommend setting this environment variable in ``/etc/ood/config/nginx_stage.
...
 
-This will control what host the shell app ssh's to when the URL accessed is ``/pun/sys/shell/ssh/default`` which is the URL other apps will use (unless there is context to specify the cluster to ssh to).
+This will control what host the shell app ssh's to when the URL accessed is ``/pun/sys/shell/`` to ``/pun/sys/shell/ssh/default`` which is the URL other apps will use (unless there is context to specify the cluster to ssh to).
+
+Set SSH Host Allowlist
+----------------------
+
+In ``/etc/ood/config/apps/shell/env`` set the env var ``OOD_SSHHOST_ALLOWLIST`` to specify what hosts can be accessed via ssh through the shell app. The format is ``login1.node.edu:login2.node.edu:*.node.edu`` and it supports wildcards. The login hosts specified in the cluster configs and the default ssh host will automatically be added to the allowlist.
+
+This will prevent the user from ssh to a host that is not specified in the allowlist.
```
ericfranz, line 264: This will prevent the user from using the Shell app to ssh to a host that is not specified in the allowlist.

source/customization.rst
ericfranz: Yeah, call out that in OnDemand 1.8 and later this will no longer default to localhost.

ericfranz: Also the sentences don't read well. Maybe:

>In ``/etc/ood/config/apps/shell/env`` set the env var ``OOD_DEFAULT_SSHHOST`` to specify the default ssh host. If not set, the default login host will be the that of the first cluster config file that has ``default: true`` in the login section. If no cluster config specifies ``default: true``, the first cluster config that does not have ``hidden: true`` in the meta section be used.

ericfranz:
This is out of date - Jeff's merge actually updated both Set Default SSH Host and Set SSH Allowlist sections appropriately.

PR #317, opened 2020-07-22
Add Turbolinks under Customization

matthu017:


johrstrom:
I think we should also note the issue with jquery's `window.onload` here.  As I recall, if folks rely on that in their `form.js`, this can break that.  

Here specifically is the context we need to add. A workaround for `window.onload`. 
https://github.com/OSC/ondemand/pull/498#issuecomment-642060505

msquee:
@johrstrom I'll start exploring options to remove jQuery as a dependency for `bc_js_filepicker`

johrstrom:
Does this impact the filepicker too?  I looked at it and it uses the event `document.ready` which I'm now seeing is a jquery event.

Weak, well I guess it could listen to both for backward comparability. In either case, this being the documentation, it should reflect whatever the truth is - which is you'd have to use file picker version 0.5.0+ if you want to enable this? 

(of course, that version doesn't exist yet, but it will by the time we merge and release)

ericfranz:
I'm going to close this right now. I think that we should first enable this feature at OSC and test it out, really understand the ramifications - places where this will definitely break (maybe that is only in the interactive app form.js stuff) and then share it in a future release.

PR #321, opened 2020-07-22
Describe pre-installed gems reduce the need for bundle install in simple apps

matthu017:
https://osc.github.io/ood-documentation/master/app-development/tutorials-passenger-apps.html

### DIFF
--- a/source/app-development/tutorials-passenger-apps.rst
+++ b/source/app-development/tutorials-passenger-apps.rst
@@ -5,7 +5,7 @@ Tutorials: Passenger Apps
 
 A Passenger App is any Rack-based Ruby app, WSGI-based Python app, or Node.js app that has a “startup file” following `the convention <https://www.phusionpassenger.com/library/config/nginx/reference/#passenger_startup_file>`__ that Phusion Passenger Web Application Server instances use to start the process. OnDemand uses the `NGINX Integration mode <https://www.phusionpassenger.com/library/walkthroughs/basics/ruby/fundamental_concepts.html#multiple-integration-modes>`__ for Passenger, where NGINX and Passenger and individual web app processes all communicate using UNIX domain sockets.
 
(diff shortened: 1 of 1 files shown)


source/app-development/tutorials-passenger-apps.rst
```diff
@@ -5,7 +5,7 @@ Tutorials: Passenger Apps
 
 A Passenger App is any Rack-based Ruby app, WSGI-based Python app, or Node.js app that has a “startup file” following `the convention <https://www.phusionpassenger.com/library/config/nginx/reference/#passenger_startup_file>`__ that Phusion Passenger Web Application Server instances use to start the process. OnDemand uses the `NGINX Integration mode <https://www.phusionpassenger.com/library/walkthroughs/basics/ruby/fundamental_concepts.html#multiple-integration-modes>`__ for Passenger, where NGINX and Passenger and individual web app processes all communicate using UNIX domain sockets.
 
-Below is a list of tutorials for developing Passenger apps for OnDemand.
+Below is a list of tutorials for developing Passenger apps for OnDemand. In addition, pre-installed gems make it easy for users to develop simple apps.
```
ericfranz, line 8: Add a second paragraph after the first that says something like:

>In addition, pre-installed gems make it easy for users to develop simple apps using Ruby. All the gems used by OnDemand are available, and in addition these gems have been added to support simple apps:

Then list the gems we added (see the Dashboard Gemfile at the bottom).

This is copy from the hpc-toolset-tutorial:

>Sinatre gem is included in gem set already available with the ondemand deployment. The ondemand gem rpms are separate rpms with version in the name so they stick around until you remove it - no loss of dependencies due to yum update. See ondemand-gems rpms at https://yum.osc.edu/ondemand/latest/web/el7/x86_64/

Also explain on the web host you can `source scl_source enable ondemand` and then `gem list` to list all the gems available on the system.

PR #327, opened 2020-07-31
Shows how to add multiple clusters on bc apps

matthu017:
This PR modifies:
https://osc.github.io/ood-documentation/master/app-development/interactive/form.html#configuration
It details how to allow bc apps to be submitted onto multiple clusters.

However this may fit under customization because it is more of an 'extra' feature rather that a necessary one.
https://osc.github.io/ood-documentation/master/customization.html

### DIFF
--- a/source/app-development/interactive/form.rst
+++ b/source/app-development/interactive/form.rst
@@ -90,12 +90,21 @@ part of the app developer can look like:
 
    # ${HOME}/ondemand/dev/my_app/form.yml
    ---
-   cluster: "owens"
+   cluster: 
+     - "owens"
+     - "pitzer"
    form:
      - bc_queue
      - bc_account
 
 The most commonly used predefined attributes are given as:
+
+cluster
+  This adds a ``select`` to the HTML form that will allow the user to select 
+  what cluster they want to submit the job.
+
+  If there is only one cluster set in ``form.yml``, the select box will not
+  be added and the job will be sent to the specified cluster.
 
 bc_account
   This adds a ``text_field`` to the HTML form that will be used as the charged
--- a/source/app-development/interactive/form.rst
+++ b/source/app-development/interactive/form.rst
@@ -90,9 +90,7 @@ part of the app developer can look like:
 
    # ${HOME}/ondemand/dev/my_app/form.yml
    ---
-   cluster: 
-     - "owens"
-     - "pitzer"
+   cluster: "owens"
    form:
      - bc_queue
      - bc_account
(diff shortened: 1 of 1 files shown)


source/app-development/interactive/form.rst
```diff
@@ -90,13 +90,22 @@ part of the app developer can look like:
...
      - bc_account
 
 The most commonly used predefined attributes are given as:
 
+cluster
+  This adds a ``select`` to the HTML form that will allow the user to select 
+  what cluster they want to submit the job.
+
+  If there is only one cluster set in ``form.yml``, the select box will not
+  be added and the job will be sent to the specified cluster.
```
ericfranz, line 107: Leave it as `cluster: "owens"`. Update the description that says cluster specifies the cluster the job will be submitted to. Then explain this can be a list of items and show a snippet below and explain what will happen if that is a list (a dropdown with both will appear). Then explain that you can use glob style match of the clusters, so could do "*" or "cluster*" if you had 2 clusters "cluster1" and "cluster2" defined in the cluster configs, they would both appear in the dropdown to select.

ericfranz: Also say this is now an OPTIONAL field as you can now specify cluster in the submit.yml instead.

source/app-development/interactive/form.rst
```diff
@@ -97,6 +97,20 @@ part of the app developer can look like:
 
 The most commonly used predefined attributes are given as:
 
+cluster
+  This specifies the cluster the job will be submitted to and is an optional field as the cluster can now be specified in the submit.yml.
+
+  This field can also be a list of items as shown below. A dropdown will appear of those options. Glob style matches are also supported - 
+  if there were two clusters named "cluster1" and "cluster2", an input of  "" or "cluster" would include both clusters in the dropdown.
```
ericfranz, line 104: A value of "*" or "cluster*"... missing the *

source/app-development/interactive/form.rst
```diff
@@ -97,6 +97,20 @@ part of the app developer can look like:
...
+
+.. code-block:: yaml
```
ericfranz, line 106: this code block should be indented the same as the text above. So:

```
  This field can also be a list of items as shown below. A dropdown will appear of those options. Glob style matches are also supported - 
  if there were two clusters named "cluster1" and "cluster2", an input of  "" or "cluster" would include both clusters in the dropdown.

  .. code-block:: yaml

     ---
     cluster:
       - "owens"
       - "pitzer"
```

* notice the yaml block starts indented to match the `c` of `code-block`

johrstrom:
I ended up doing this in this PR in this commit just because there's a lot of context around all these different options that aren't immediately apparent.

https://github.com/OSC/ood-documentation/pull/364/commits/ea9699b5e38c35b415360d0ac411aa84188a0ae2

PR #376, opened 2020-08-18
Use SVG logo

msquee:
https://osc.github.io/ood-documentation-test/svg-logo/

**SVG**

**Alignment:**

vs.

**PNG**


ericfranz:
revisit after we make a decision about switching to docusaurus

PR #377, opened 2020-08-18
Add template reference

johrstrom:
https://osc.github.io/ood-documentation-test/add-template-reference/

Now that we have reference, I can finally make a PR for the other docs I had :laughing: .  

I'm trying to PR this into develop just because it doesn't need to be rushed into the 1.8 release.


johrstrom:
Let's open after 1.8 is released, this has a stupid amount of differences.

PR #381, opened 2020-08-19
Add more release notes

msquee:
Add more release notes:
* IE 11 drop
* Sinatra gems
* Copy and paste fix on Chrome noVNC

https://osc.github.io/ood-documentation-test/even-more-rns/release-notes/v1.8-release-notes.html#drop-ie-11-support

Need to go back and update the reference link to `Browser Requirements` once it's available upstream.

### DIFF
--- a/source/release-notes/v1.8-release-notes.rst
+++ b/source/release-notes/v1.8-release-notes.rst
@@ -20,6 +20,9 @@ Highlights in 1.8:
 - `Support for SVG dashboard logos`_
 - `Support for new Native VNC views`_
 - `Better max file upload support`_
+- `Drop IE 11 support`_
+- `Added Sinatra RubyGems into ondemand-gems for other apps to use`_
+- `Streamlined copy and paste in noVNC for Chrome`_
 
 Special thanks
 --------------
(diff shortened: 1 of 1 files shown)


source/release-notes/v1.8-release-notes.rst
```diff
@@ -257,3 +260,28 @@ nginx configuration so that sites can set the upload max to settings larger than
...
+.. code-block:: ruby
+
+  gem "sinatra", require: false
+  gem "sinatra-contrib", require: false
+  gem "erubi", require: false
+
+Streamlined copy and paste in noVNC for Chrome
+..............................................
+
+There was an issue with copy and pasting large clipboard buffers into noVNC applications on Chrome. This was fixed in version 1.8.
```
johrstrom, line 287: Isn't this feature just that copy/paste to the clipboard just works in Chrome? I mean there _was_ this bug too, but the real value add here is that copy/paste works out of the box for VNC apps instead of having to use the NoVNC workaround.

msquee: Fixed

johrstrom:
Works for me!

PR #384, opened 2020-08-20
Xdmod and release notes updates

ericfranz:
https://osc.github.io/ood-documentation-test/xdmod_and_release_notes_updates/


source/customization.rst
```diff
@@ -1,4 +1,4 @@
-.. _customization:
+...................... _customization:
```
johrstrom, line 1: Is this some typo? It shows up as the literal `…………………. _customization:`

source/customization.rst
```diff
@@ -966,3 +966,54 @@ Changes are applied by running ``update_ood_portal`` and restarting the ``ondema
...
+   .. code-block:: none
+
+      domains = "https://ondemand.osc.edu"
+
+#. Configure identity provider to include OnDemand host in HTTP Content-Security-Policy for frame-ancestors since OnDemand uses iFrames to trigger SSO with XDMoD when a user logs in
+
+   .. code-block:: none
+
+      frame-ancestors https://*.osc.edu 'self'
+
```
johrstrom, line 1002: Is this code example for KeyCloak or Dex?

ericfranz: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors

It is the value of the Content-Security-Policy header

It was what we had to set in Keycloak but I guess the full header would look like

    Content-Security-Policy: frame-ancestors https://*.osc.edu 'self'

source/customization.rst
ericfranz: Fixed the typo. It no longer appears in the diff.

source/customization.rst
ericfranz: I added a link to MDN and clarified the example was the value of the Content-Security-Policy header that was used in Keycloak at OSC

PR #385, opened 2020-08-20
Add template reference

johrstrom:
https://osc.github.io/ood-documentation-test/add-template-reference/

Now that develop is back up to date I can re-open this. And it's timely because we just got a question on discourse about `min_port` and `max_port` which we could point the user back to this documentation.


ericfranz:
Move the documentation on submit.yml.erb to the section along with https://osc.github.io/ood-documentation/latest/app-development/interactive/submit.html since we discuss submit.yml.erb there.

Then add reference this in "Setup Interactive Apps" (https://osc.github.io/ood-documentation/latest/app-development/interactive/setup.html) as a step "4. Configure Global Options" by saying something like "configure all of the batch connect apps to behave in a particular way by specifying configuration options globally in the cluster config to link to. Examples of what can be configured include specifying min/max ports, path to websockify, (etc. etc.) and then link to the page that documents each as you have here.

source/reference/files/submit-yml/script-options.rst
```diff
@@ -0,0 +1,460 @@
+.. _script-options:
+
+Script Options
+==============
```
ericfranz: Instead of documenting this here, add any appropriate example documentation inline in the code for ood_core and link to generated yardoc for ood_core from here.

Currently this is available from https://osc.github.io/ood_core/ but we are not auto-generating that yet for each - a travis ci or github actions needs added to do that...

ericfranz: https://osc.github.io/ood_core/0.11.4/OodCore/Job/Script.html

The point is the overview is just one string of text:

>An object that describes a batch job before it is submitted. This includes the resources this batch job will require of the resource manager.

But it could be detailed description of the use of the batch connect yaml how it corresponds to the defined methods. Or enough information so that each method documentation provides its own example of using via the batch connect yaml.

johrstrom: Won't that be confusing to users to have to interpret ruby docs?  I just mean they'll be unfamiliar with it and it may not be easy to read for them.

ericfranz: I guess it depends on how the Ruby docs are formatted. If there is enough explanation above it probably isn't that confusing. FWIW we could also use the yard tool to generate a restructured version of this documentation and then use that in here. Adding directions to help address any confusion of "intepretting ruby docs" is easiest though.

johrstrom: Maybe we can just get rid of this page so the PR can move forward? I can then create follow up tickets to (a) create a tasks to auto publish the yarn docs and (b) add a link to that here.

ericfranz: Sounds good

johrstrom:
I removed script options as we've discussed. I think this is ready now.

PR #395, opened 2020-08-28
Add HTML meta tags for Algolia DocSearch

msquee:
This PR is the scaffolding for a successful Algolia DocSearch deployment! We recently got approved for a [DocSearch](https://docsearch.algolia.com/) account since Open OnDemand is open source.

**TLDR: We do not want to return search results from version 1.0 with `latest`**

This PR needs to be merged into `latest` because Algolia scrapes `https://osc.github.io/ood-documentation/master/` every 24 hours. The JavaScript code to enable DocSearch looks like this:
```js
docsearch({
  apiKey: PUBLIC_API_KEY,
  indexName: 'osc_ood',
  inputSelector: '#rtd-search-form input[type=text]',
  debug: false,
  algoliaOptions: {
    hitsPerPage: 6,
    facetFilters: [
      'version:1.8.0'
    ]
  },
  custom_settings: {
    attributesForFaceting: ['language', 'version', 'branch']
  },
})
```

DocSearch specifically looks for meta tags named `docsearch:TAG_NAME` and adds it to the search index so results can be filtered based on tags. `facetFilters` is the important bit here, we don't have the following global meta tags deployed on `latest`:

```html
<meta name="docsearch:language" content="en"/>
<meta name="docsearch:version" content="1.8.0"/>
<meta name="docsearch:branch" content="latest"/>
```

Once those are added and Algolia re-scrapes the documentation, then I can introduce the PR that is behind the demos.

**Why we want DocSearch:**
- Free, if we go over the limits then we can self host our own DocSearch Docker image. See [DocSearch FAQ](https://docsearch.algolia.com/docs/faq/).
- Algolia is amazing, look at the screenshot below for example. Typo tolerance will be great for searches.
- Inline and instant search, look at the [Video Demo (20s)](https://streamable.com/4wxy5u).

**Screenshot:**

**Video Demo (20s):** https://streamable.com/4wxy5u

### DIFF
--- a/source/_templates/layout.html
+++ b/source/_templates/layout.html
@@ -2,6 +2,10 @@
 
 {% block extrahead %}
 {{ super() }}
+<meta name="docsearch:language" content="en"/>
+<meta name="docsearch:version" content="1.8.0"/>
+<meta name="docsearch:branch" content="latest"/>
+
 <script async src="https://www.googletagmanager.com/gtag/js?id=UA-34776817-3"></script>
 <script>
   window.dataLayer = window.dataLayer || [];


ericfranz:
The commit should be to develop and be able to be merged into all the branches. So if I have it correctly, there are two steps:

1. meta tags specify which git branch and app version a specific documentation URL is from
2. docsearch is configured/told which URLs to scrape

I don't recall why branch matters to docsearch if docsearch is just scraping the statically produced HTML. How does docsearch interact with the git repo?

I also notice version specified is 1.8.0 where the patch release is incorrect (current is 1.8.13). Can we just specify 1.8? What happens when the version string changes?

source/_templates/layout.html
```diff
@@ -2,6 +2,10 @@
 
 {% block extrahead %}
 {{ super() }}
+<meta name="docsearch:language" content="en"/>
+<meta name="docsearch:version" content="1.8.0"/>
+<meta name="docsearch:branch" content="latest"/>
```
ericfranz, line 7: These hardcode 1.8.0 and latest. Instead they should use the sphinx variables:

https://github.com/OSC/ood-documentation/blob/9d24e708aa435e8e8a908c474d005c65c9d53e09/source/conf.py#L69-L72

https://github.com/OSC/ood-documentation/blob/9d24e708aa435e8e8a908c474d005c65c9d53e09/source/conf.py#L152

ericfranz: You can search in the docs where these variables are used elsewhere.

ericfranz: If we only want the meta tags to be in latest then we should have an if block around these meta tags testing if the github_version is latest.

PR #473, opened 2021-02-03
Update enabling-development-mode.rst

rsettlage:

https://osc.github.io/ood-documentation-test/BRANCH-NAME/

**Add your description here**

### DIFF
--- a/source/app-development/enabling-development-mode.rst
+++ b/source/app-development/enabling-development-mode.rst
@@ -3,7 +3,7 @@
 Enabling App Development
 ========================
 
-Enable in OnDemand v1.6:
+Enable in OnDemand v1.6+:
 ........................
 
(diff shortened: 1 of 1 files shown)


source/app-development/enabling-development-mode.rst
```diff
@@ -3,7 +3,7 @@
 Enabling App Development
 ========================
 
-Enable in OnDemand v1.6:
+Enable in OnDemand v1.6+:
 ........................
```
johrstrom, line 7: You should be branching from develop, not latest. But even so, try to `make html` if you can (I'm not sure why the automation isn't running for a fork, but it should).  

In any case, this may fail because RST needs the dotted underline to be as long as the thing that it underlines, so you're likely going to have to extend it too, but building would tell us for sure.

PR #505, opened 2021-05-18
more detailed highlight & prioritize files app

ericfranz:
https://osc.github.io/ood-documentation-test/release_2.0_notes_changes/

- instead of having "Breaking Changes" as a highlight, just moved the warning to the top of the page
- reworded Pinned Apps and Custom widgets highlight titles to attempt to better describe what they are
- reworded section on integrating Active Jobs into the dashboard to also include the other newly integrated apps in the same section
- added missing "Upgrade directions" header after the "Breaking changes" header

Note: there are a few places that could be made links that do not. Since I don't have a lot of time instead of adding review comments on that just consider doing it!

More to  come in separate PRs. Generally, we should expand the sections of the release notes so that they at least contain descriptions as or more verbose than https://discourse.osc.edu/t/announcing-pre-release-of-open-ondemand-2-0/1450

But I think we should consider going a little further with this. For example, we could reiterate some of the goals of the features (outlined in the cards https://trello.com/b/ksr1g141/open-ondemand-ideas-and-dev) and how those are accomplished.

Maybe Buffalo would let us use their screenshot :-).

### DIFF
--- a/source/release-notes/v2.0-release-notes.rst
+++ b/source/release-notes/v2.0-release-notes.rst
@@ -3,16 +3,20 @@
 v2.0 Release Notes
 ==================
 
+.. warning::
+
+   There are some breaking changes in 2.0. See the upgrade directions below for details.
+
+
 Highlights in 2.0:
 
-- `Breaking Changes`_
-- `Pinning apps to the dashboard`_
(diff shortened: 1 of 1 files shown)


johrstrom (approved):
Fine by me!

Issue #507, opened 2021-05-18
Highlights missing: controlling new window or same window link behavior


ericfranz:
Apps can now define the open in new window behavior through their manifest https://github.com/OSC/ondemand/pull/1094

johrstrom:
This got fixed in #518.

PR #518, opened 2021-05-18
open link in same window and dalli highlights

### LINKED ISSUE #508: Highlight missing: dalli gem
dalli is now an optional ruby gem that can be required and used in interactive apps (or custom initializers if using memcachd for any reason - maybe link to original discourse topic

ericfranz:
Fixes #508 and #507


johrstrom (approved):
LGTM

PR #536, opened 2021-07-02
use latest instead of master as default

johrstrom:
The root of our GH pages uses `master` which we don't update and in fact need to get rid of. Instead we should be using latest.

This will change the behaviour when you land here:
https://osc.github.io/ood-documentation/

And you _should_ be redirected here:
https://osc.github.io/ood-documentation/latest


Oglopf (approved):
LTGM.

PR #544, opened 2021-09-01
cacheable should be false in example

johrstrom:
https://osc.github.io/ood-documentation-test/fix-cacheable-example/

Fixes this example to show that cacheable is false.


Oglopf (approved):
LTGM.

PR #558, opened 2021-10-21
add docs for the pun_pre_hook

### LINKED ISSUE #439: OOD PR add pun_pre_hook (#535)
Create documentation for https://github.com/OSC/ondemand/commit/28da29dc512cadc14a29e33685c60c8446241d36

johrstrom:
https://osc.github.io/ood-documentation-test/add-pun-pre-hook

Fixes #439.  I'm going to create a new ticket to move the docs about the `hook.env` out of the k8s cluster config doc.

### DIFF
--- a/source/installation/resource-manager/kubernetes.rst
+++ b/source/installation/resource-manager/kubernetes.rst
@@ -114,6 +114,7 @@ You'll need to employ :ref:`PUN pre hooks <ood-portal-generator-pun-pre-hook>`
 to bootstrap your users to this cluster. You'll also have to modify ``/etc/ood/config/hooks.env``
 
 .. code-block:: text
+
   # /etc/ood/config/hook.env
 
   # required if you changed the items in the cluster.d file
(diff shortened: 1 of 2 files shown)


source/installation/resource-manager/kubernetes.rst
```diff
@@ -110,6 +110,32 @@ The RoleBindings give user, ``oakley`` in this case, sufficient privileges
...
+  CLIENT_ID="changeme"
+  CLIENT_SECRET="changeme"
+
+  # required if you're using a secret registry
+  IMAGE_PULL_SECRET=""
+  REGISTRY_DOCKER_CONFIG_JSON="/some/path/to/docker/config.json"
+
+  # enable if are enforcing walltimes through the job pod reaper
+  # see 'Enforcing walltimes' below.
+  USE_JOB_POD_REAPER=false
```
johrstrom: This isn't really the place for this, but it'll have to work for now.

treydock:
So the one piece I see missing is what we deploy here: https://github.com/OSC/osc-ood-config/blob/master/hooks/pre-hook.sh.  That hook does call some scripts that are shipped with OnDemand, but the pre-hook.sh is not shipped with OnDemand.  Might be good for us to include an example in main OnDemand repo like we do with hook.env, here: https://github.com/OSC/ondemand/tree/master/hooks.  I think maybe having an example in the source code and referencing that example so we can be like "this is what a hook might look like if you want to use scripts we provide".

johrstrom:
Added the link and the full text. Still not in the right place (maybe even not the right order). In fact, we'd probably just need to add docs for all the scripts individually too.

PR #596, opened 2021-11-09
Adding web and compute node definitions with examples.

### LINKED ISSUE #595: Define explicitly the web-node and compute-node
When I came on board I found the use of these terms confusing and was never able to find clear definitions of what these are in the docs.

I've noticed in user groups and discourse that there seems to also be confusion among users about what exactly these terms mean. It would really be nice to have a clear, up front definition of what these things are and what they do so that users aren't so confused trying to keep track of all the moving pieces.

We do call out what the web-node is in the architecture page, but I can't find the compute-node really being discussed as a term, which can be very frustrating as a user.

Oglopf:

https://osc.github.io/ood-documentation-test/compute-web-node-defns/

altered `architecture` page.

**Add your description here**
Adding explicit definitions for the web-node and compute-node in the overview. Fixes #595

### DIFF
--- a/source/architecture.rst
+++ b/source/architecture.rst
@@ -10,13 +10,29 @@ Below are some diagrams of OnDemand's architecture:
    model for software diagrams, are more technically detailed and are built using draw.io
 #. Request flow diagram is a sequence diagram built using plantuml
 
+Useful Terms
+------------
+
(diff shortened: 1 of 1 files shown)


Oglopf:
Closing this PR out to open a separate PR to add a glossary for terms.

PR #597, opened 2021-11-09
Adding glossary and some grammar cleaning.

### LINKED ISSUE #595: Define explicitly the web-node and compute-node

Oglopf:

https://osc.github.io/ood-documentation-test/glossary/

Adding `glossary`, adjusted `architecture` and `index`.

**Add your description here**
Adding a glossary page to start collecting terms. Fixed some missing periods and adjusted some wording on the architecture page. Fixes #595

### DIFF
--- a/source/glossary.rst
+++ b/source/glossary.rst
@@ -4,7 +4,7 @@ Glossary
 ========
 
    Cluster
-     Logical set of physical machines.
+     Logical set of physical machines with a resource scheduler that users can submit jobs to.
 
    Compute-node
      Where the actual compute happens using information passed from a web-node or submitted scripts. Generally these nodes:
--- a/source/glossary.rst
+++ b/source/glossary.rst
@@ -1,0 +1,32 @@
+.. _glossary:
+
+Glossary
+========
+
+   Cluster
+     Logical set of physical machines.
+
+   Compute-node
+     Where the actual compute happens using information passed from a web-node or submitted scripts. Generally these nodes:
+    
+       * Run as authenticated user.
+       * Can use the web-node forms for data to start a job.
+       * Can use submitted scripts from the shell or job-composer for resource managers like SLURM, Torque, etc.
+       * The Per-User NGINX serves web apps in Ruby and NodeJS and is how users submit jobs and start interactive apps.
+  
+   Login-node
+     Allows you to interact with a compute node live to submit computations and see the results in a REPL fashion with examples like:
+  
+     * Jupyter
+     * Rstudio
+     * Ansys
+
+   Web-node
+     Refers to the Apache web server front-end:
+      
+      * Authenticates user.
+      * Runs as apache user.
+      * Handles authentication and tracking the user session.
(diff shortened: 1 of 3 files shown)


Oglopf:
The ``:term:`<term_to_ref>` `` functionality was not working for me out of the box. I'd like to still add that at some point when we have some time to see why it won't reference correctly.

source/glossary.rst
```diff
@@ -0,0 +1,32 @@
+.. _glossary:
+
+Glossary
+========
+
+   Cluster
+     Logical set of physical machines.
```
johrstrom:
```suggestion
   Cluster
     Logical set of physical machines with a resource scheduler that users can submit jobs to.
```

source/glossary.rst
```diff
@@ -0,0 +1,32 @@
...
+
+   Compute-node
+     Where the actual compute happens using information passed from a web-node or submitted scripts. Generally these nodes:
+    
+       * Run as authenticated user.
+       * Can use the web-node forms for data to start a job.
+       * Can use submitted scripts from the shell or job-composer for resource managers like SLURM, Torque, etc.
+       * The Per-User NGINX serves web apps in Ruby and NodeJS and is how users submit jobs and start interactive apps.
```
johrstrom:
```suggestion
   Compute-node
     Jobs run on compute nodes.
```

There could be more here - but I think we need to side on conciseness.

source/glossary.rst
```diff
@@ -0,0 +1,32 @@
...
+
+   Web-node
+     Refers to the Apache web server front-end:
+      
+      * Authenticates user.
+      * Runs as apache user.
+      * Handles authentication and tracking the user session.
+      * Tracks the various `fs` paths that it will associate with the compute nodes which it submits requests to using form data.
+      * Reverse proxies each user to her PUN via Unix domain sockets.
+      * Reverse proxies to interactive apps running on compute nodes (RStudio, Jupyter, VNC desktop) via TCP sockets.
```
johrstrom:
```suggestion
   Web-node
     Generically Web-Node refers to a machine that has web applications running on it.
     
     Throughout these documents it refers to the machine that Open OnDemand is installed on.
```

johrstrom:
I left just 2 more suggestions. Node is a synonym for machine. Apache and PUNs both run on the web node, but those processes aren't themselves nodes. They're just apps running on a node.

Oglopf:
Let me know if those definitions look any better. 

I added the web-login as well.

johrstrom (approved):
Thanks!

PR #606, opened 2021-11-30
use cpu_limit instead of duplicate cpu_request

johrstrom:
Just a quick fix `cpu_limit` wasn't correctly documented.  This is going straight into latest.

### DIFF
--- a/source/app-development/tutorials-interactive-apps/k8s-jupyter.rst
+++ b/source/app-development/tutorials-interactive-apps/k8s-jupyter.rst
@@ -76,7 +76,7 @@ Kubernetes has some flexibility in requests. One can make _requests_ and _limits
 which are like hard and soft limits. In the example above, they're both the same.
 
 Here's an example utilizing requests and limits for both memory and cpu. Note that
-we're using millicores in the ``cpu_request``.
+we're using millicores in ``cpu_request``.
 
 .. code-block:: yaml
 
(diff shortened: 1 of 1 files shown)


source/app-development/tutorials-interactive-apps/k8s-jupyter.rst
```diff
@@ -76,15 +76,15 @@ Kubernetes has some flexibility in requests. One can make _requests_ and _limits
...
 Here's an example utilizing requests and limits for both memory and cpu. Note that
-we're using millicores in the ``cpu_request``.
+we're using millicores in ``cpu_request``.
 
 .. code-block:: yaml
 
     container:
       # ...
       port: "8080"
       cpu_request: "0.200"
```
treydock, line 86: A lot of Kubernetes docs uses `200m` rather than `0.200`. Wonder if we should use that same millicore unit.  Example: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu

treydock: I personally like `0.200` more but that's because to me `m` requires too much context and knowledge of what it means to Kubernetes.  Just a thought in case we wanted to make our docs look like Kubernetes.

johrstrom: Yea I even used `m` in our apps. I could go either way - I feel like if you run Kubernetes and/or am reading these docs you're aware of this stuff?

PR #614, opened 2021-12-09
Fix alternate text cluster selection widget example

novosirj:
The ordering of the cluster name and the alternate text was incorrect in the provided example. It works properly the other way around.

https://osc.github.io/ood-documentation/latest/app-development/interactive/form.html?highlight=cluster#configuring-which-cluster-to-submit-to

### DIFF
--- a/source/app-development/interactive/form.rst
+++ b/source/app-development/interactive/form.rst
@@ -602,8 +602,8 @@ different text than the default.
      cluster:
        widget: "select"
        options:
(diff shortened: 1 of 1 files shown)


novosirj:
Apologies if I've done this incorrectly. I do pull requests juuuust infrequently enough to not be confident on how to do them. 😅

johrstrom:
Nope! That's just fine.

PR #619, opened 2022-01-12
Request flow diagram page added for better user understanding of onde…

Oglopf:
…mand request handling.

https://osc.github.io/ood-documentation-test/add-app-flow-diagrams/

**Add your description here**
Adding some diagrams about the different request flows using various apps in OnDemand.


source/index.rst
```diff
@@ -51,6 +51,7 @@ These are institutions who were early adopters or provided HPC resources for dev
    :caption: General
 
    architecture
+   request-flow-diagrams
```
johrstrom: I want to be cautious in adding new headings here. Indeed I'll be making pull request to reduce them.

It's fine to be it's own page, but would this make more sense as a subheading within architecture (or some other heading)?

Oglopf: I think that makes a lot more sense. I went ahead and created a subheading for these under architecture which reads much cleaner.

PR #621, opened 2022-01-19
update user mapping documentation

johrstrom:
https://osc.github.io/ood-documentation-test/update-user-mapping/
and specifically - 
https://osc.github.io/ood-documentation-test/update-user-mapping/authentication/overview/map-user.html

This corrects the documentation on user mapping to be consistent with 2.0. As such it's going directly into latest.  It removes some stuff about the `.regex` file we used to provides and add some things like debugging tips and a note about dex.

### DIFF
--- a/source/authentication/overview/map-user.rst
+++ b/source/authentication/overview/map-user.rst
@@ -3,99 +3,114 @@
 Setup User Mapping
 ==================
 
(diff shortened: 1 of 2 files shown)


Oglopf (approved):
LGTM.

PR #623, opened 2022-01-27
Moving app dev and interactive apps tutorial under how-tos section.

Oglopf:

https://osc.github.io/ood-documentation-test/move-intrct-apps-to-how-to/

**Add your description here**
Cleaning up the docs index my moving more of the development and tutorial into subsections under the How-To section for easier user navigation.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1201730403685694) by [Unito](https://www.unito.io)


source/index.rst
```diff
@@ -75,13 +75,15 @@ These are institutions who were early adopters or provided HPC resources for dev
...
+  how-tos/app-development
+  how-tos/tutorials-interactive-apps
 
 
 .. toctree::
    :maxdepth: 2
    :caption: Extend
 
-   app-development/interactive/setup
+   how-tos/app-development/interactive/setup
```
johrstrom, line 91: just remove this link if it's a child of `how-tos/app-development`.

source/index.rst
```diff
@@ -75,13 +75,15 @@ These are institutions who were early adopters or provided HPC resources for dev
   :caption: How-Tos
 
   how-tos/apps.rst
+  how-tos/app-development
+  how-tos/tutorials-interactive-apps
```
johrstrom: Actually - Now that I see this, seems like there should be `How Tos` and `Tutorials` sections with `tutorials-interactive-apps` moving into `Tutorials`.

Oglopf: I kind of like just having the tutorials under the How-Tos, since a tutorial is just showing you "how to do something" it seems like it makes sense to me. It also prevents that TOC from becoming cluttered with sections. I'd find it kind of odd to be a user and see a "how to" and then a "tutorial" section, it just seems semantically redundant.

I'm happy either way though, and regardless the passenger tutorial should also be moved either into the Tutorials if we go that route, or also under the "How Tos".

source/index.rst
Oglopf: This one has me scratching my head. Removing it breaks the build.  

For now I'd rather leave this and handle refactoring this section separately as its own issue since it was not in scope for this issue. Plus, moving that will be tricky for me because the toctrees are so...interesting.

Basically, let's make another ticket to refactor this section as well because the toctree is confusing and needs to be decyphered and fixed.

source/index.rst
Oglopf: Broke out into the Tutorials and How-Tos sections and put the passenger and interactive app tutorials under the former and the app development under the latter.

johrstrom (approved):
LGTM

PR #630, opened 2022-01-28
shift the main sections around adding a getting started

johrstrom:
https://osc.github.io/ood-documentation-test/getting-started/

Shift the main sections around adding a getting started. This doesn't modify any of the underlying files, just moves things around.

I think we should only have these 4 sections (Legacy docs didn't move, we just need to deal with them directly).

```
Getting Started
How-Tos
Tutorials
More
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737269114337) by [Unito](https://www.unito.io)

### DIFF
--- a/source/index.rst
+++ b/source/index.rst
@@ -48,30 +48,27 @@ These are institutions who were early adopters or provided HPC resources for dev
 
 .. toctree::
    :maxdepth: 2
-   :caption: General
-
-   architecture
-   reference
-   release-notes
-   glossary
-
-
-.. toctree::
-   :maxdepth: 2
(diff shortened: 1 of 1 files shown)


Oglopf:
I really like this layout after using it because it was much more intuitive to find information under those new sections. 

Whether something like the "Customization Overview" goes in the Getting Started or How-Tos under the Customization subsection there is something we can talk about, but the general 4 sections listed I really like for navigation.

Oglopf (approved):
LGTM.

PR #637, opened 2022-02-09
k8s docs should say host not server

johrstrom:
https://osc.github.io/ood-documentation-test/johrstrom-patch-1/

k8s docs should say host not server


Oglopf (approved):
LGTM.

johrstrom:
Thanks. I reset it back to just the 1 update and opened a ticket for `sub_path`s which I'd added just then, but then removed because it's a slightly bigger change than what I can patch inline here.

PR #638, opened 2022-02-09
Removing index reference, from-source page, and from-source directory.

### LINKED ISSUE #625: Build From Source Removal
We need to remove the build from source instructions as it has lead several users in the open office hours to reveal they've experienced some pain by trying to follow those docs instead of the RPM install.

Let's remove this page as it is of no use to the end users, is essentially noise, and it's questionable how out of date some of it is.

Also, collect the other "source" issues to update those pages and just remove those issues when this is implemented.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737349128553) by [Unito](https://www.unito.io)

Oglopf:

https://osc.github.io/ood-documentation-test/remove-source-install/

**Add your description here**
Fixes #625

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201795342787456) by [Unito](https://www.unito.io)


johrstrom:
Can you update the installation.rst to mention something along the lines of

```
Building from source is left as an exercise to the reader. It's not particularly difficult to build the code, but installing it with all the various files is. Should you be interested, review the Dockerfile and packaging specs for what would be involved.

If you'd like a package built for a system that we don't currently support, feel free to open a ticket!

https://github.com/OSC/ondemand/blob/master/Dockerfile
https://github.com/OSC/ondemand/tree/master/packaging
https://github.com/OSC/ondemand/issues/new
```

source/installation.rst
```diff
@@ -18,13 +18,26 @@ means that it will need:
 - your LDAP URL, base DN, and attribute to search for (in some rare cases a
   bind DN and corresponding bind password)
 
+Building From Source
+--------------------
```
johrstrom: Can this section be under the TOC tree? Let's promote the more common use case.

Oglopf: Made a change to move this to the last item in the TOC tree for installs. Let me know if that looks like what you were thinking.

johrstrom (approved):
LGTM

PR #643, opened 2022-02-18
Update slurm config documentation to reference additional commands if not using SCL

gbyrket:
https://osc.github.io/ood-documentation-test/updateSlurmDoc/installation/resource-manager/test

Update slurm config documentation to reference additional commands if not using SCL

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201852664276197) by [Unito](https://www.unito.io)

### DIFF
--- a/source/installation/resource-manager/test.rst
+++ b/source/installation/resource-manager/test.rst
@@ -22,7 +22,7 @@ configuration files.
 
 #. We will now list all available tasks that we can run:
 
-   .. rubric:: If you are using SCL, run the following command:
+   .. rubric:: If your O/S is CentOS 7 or RHEL 7, run this command:
 
    .. code-block:: sh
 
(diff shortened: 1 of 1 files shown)


johrstrom:
Can the language center more around the OS you're on?  SCL is only centos:7 so really it's `centos:7`  vs every other platform.

Seeing phrases like `using SCL` and `NOT using SCL` is a bit hard to distinguish - plus how would they know? 

I think phrasing it as what operating system you're on may be more understandable.

```
if you're on CentOS/RHEL 7 do this
...
otherwise do that.
...
```

source/installation/resource-manager/test.rst
```diff
@@ -22,7 +22,7 @@ configuration files.
 
 #. We will now list all available tasks that we can run:
 
-   .. rubric:: If you are using SCL, run the following command:
+   .. rubric:: If your O/S is CentOS 7 or RHEL 7, run this command:
```
johrstrom:
```suggestion
   .. rubric:: If your operating system is CentOS 7 or RHEL 7, run this command:
```

Let's try not to use abbreviations.

johrstrom (approved):
LGTM

Issue #636, opened 2022-02-07
Get rid and/or move legacy docs


johrstrom:
Get rid and/or move legacy docs. They'll always be in git, so it's fine to remove.  I've only just glanced at them now, so I don't know how useful they are, but some portions could be useful.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201781146331178) by [Unito](https://www.unito.io)

johrstrom:
We did this on the develop branch at some point.

PR #644, opened 2022-02-18
rm some references to legacy docs

johrstrom:
https://osc.github.io/ood-documentation-test/rm-legacy-db/

This reference to this legacy doc is all over the place here. In an effort to limit the changes in a PR - I've just changed a few occurrences of it. There are still more. Then, when all references are removed can we take care of #636.

```
:ref:`dashboard`
````


gbyrket (approved):
LGTM

PR #646, opened 2022-02-21
make install as source a 2nd header

johrstrom:
https://osc.github.io/ood-documentation-test/rm-source/

Because this section had a ==== it was the interpreted as having the same section heading as the rest of authentication and thus a part of the menu bar here. This makes it so it's not a part of the menu bar.

image


Oglopf (approved):
Good catch. LGTM!

PR #657, opened 2022-03-17
typo: websockify_cmd

jose-d:
I believe this was only typo appearing here:

https://osc.github.io/ood-documentation/latest/reference/files/submit-yml/vnc-bc-options.html?highlight=websockfiy_cmd

reproducer:

1) go to the URL above
2) search for string "websockfiy_cmd"

IMO `websockfiy_cmd` should be `websockify_cmd`.


johrstrom (approved):
Thanks! You're right. I had to look it up to be sure we didn't mispell the original!****

PR #656, opened 2022-03-16
mv customization overview to reference

johrstrom:
**Modify this link to include the branch name, and possibly the page this PR modifies**:

https://osc.github.io/ood-documentation-test/coalese-customize/

This moves the `customization overview` to reference and makes `customizations` a direct navigation link.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201981671630881) by [Unito](https://www.unito.io)

PR #659, opened 2022-03-18
mv customization overview page to highlight customizing itself

johrstrom:

https://osc.github.io/ood-documentation-test/mv-customize/

Same as #656 to move the customization overview to reference and makes customizations a direct navigation link.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201993213319603) by [Unito](https://www.unito.io)


Oglopf:
This looks good and I like breaking that PUN section out. Should we come up with a better section header than "More" at this point though? We can handle that later, but it seems like some great resources are down in that section now and it may not catch a user's eye when navigating the docs is all.

Oglopf (approved):
LGTM.

johrstrom:
> Should we come up with a better section header than "More" at this point though?

Yes I think so

PR #666, opened 2022-04-14
move dynamic bc info and give an example for form yml

johrstrom:
https://osc.github.io/ood-documentation-test/interactive-updates/

Move dynamic bc info within interactive apps. This also gives an example/overview of the `form.yml` which I think is a bit complicated and could use some helpers around what all these files do and how they relate & interact with each other.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202132105045427) by [Unito](https://www.unito.io)


gbyrket (approved):
Each page that you modified looks good.

PR #671, opened 2022-04-26
install instructions to use tabs

### LINKED ISSUE #667: use tabs to toggle OS flavors in install instructions
I was just working through the install instructions and wondering how to restructure it.

Looking at pegasus' documentation we can have these tabs which are much nicer.

https://pegasus.isi.edu/documentation/user-guide/installation.html#rhel-centos-sl

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202158386666511) by [Unito](https://www.unito.io)

johrstrom:
Fixes #667. 

https://osc.github.io/ood-documentation-test/tab-install-instructions/

This add tabs so that the install instructions are much more readable when determining your OS.

Navigate to the install instructions and this is what you'll see.
image

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202184358598762) by [Unito](https://www.unito.io)

### DIFF
--- a/source/installation/install-software.rst
+++ b/source/installation/install-software.rst
@@ -16,76 +16,80 @@ software requirements:
    This tutorial is run from the perspective of an account that has
    :command:`sudo` access but is not root.
 
-#. Enable the dependency repositories **on CentOS/RHEL 7 only**:
+1. Enable the dependency repositories
 
-   CentOS 7
-     .. code-block:: sh
+.. tabs::
 
-        sudo yum install centos-release-scl epel-release
+   .. tab:: CentOS 7
 
-   RHEL 7
-     .. code-block:: sh
+      .. code-block:: sh
 
-        sudo yum install epel-release
-        sudo subscription-manager repos --enable=rhel-server-rhscl-7-rpms
-        # Repository 'rhel-server-rhscl-7-rpms' is enabled for this system.
+         sudo yum install centos-release-scl epel-release
 
-   .. warning::
 
-      For **RedHat** you may also need to enable the *Optional* channel and
-      attach a subscription providing access to RHSCL to be able to use this
-      repository.
+   .. tab:: RockyLinux 8
 
-#. Enable dnf modules and repositories for dependencies **on RHEL/Rocky 8 only**:
+      .. code-block:: sh
 
-    .. code-block:: sh
+         sudo dnf config-manager --set-enabled powertools
+         sudo dnf install epel-release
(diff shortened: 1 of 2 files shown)


source/installation/install-software.rst
```diff
@@ -16,76 +16,80 @@ software requirements:
...
+2. Add Open OnDemand's repository and install:
 
-       sudo apt install apt-transport-https ca-certificates
-       wget -O /tmp/ondemand-release-web_{{ ondemand_version }}.0_all.deb https://apt.osc.edu/ondemand/{{ ondemand_version }}/ondemand-release-web_{{ ondemand_version }}.0_all.deb
-       sudo apt install /tmp/ondemand-release-web_{{ ondemand_version }}.0_all.deb
-       sudo apt update
+   .. tabs::
 
-#. Install OnDemand and all of its dependencies:
+      .. tab:: RHEL clones
```
johrstrom: I don't know if this is the right phrase for RHEL type systems.

treydock: I see some places refer to RHEL and RHEL clones as "Enterprise Linux" since RHEL stands for "Red Hat Enterprise Linux".

treydock:
Looks like missing tabs for installing ondemand-dex since there is EL vs Ubuntu steps.

source/installation/install-software.rst
treydock: I also see some documentation refer by package manager so I will see like `yum/dnf` and `apt` as ways to describe sections.

johrstrom:
> Looks like missing tabs for installing ondemand-dex since there is EL vs Ubuntu steps.

I'm deferring that because i want to restructure optional things anyhow.  Maybe even until https://github.com/OSC/ondemand/issues/1965

source/installation/install-software.rst
johrstrom: I was thinking the same thing about sudo, but I feel like we have to be super explicit or we'll get topics on discourse like 'yum/dnf command not found' lol.

treydock: Oh I meant the tab could say `yum/dnf` instead of `RHEL clones` but the actual commands would all still use `yum` because that exists still in RHEL8 but might be gone in RHEL9, not sure yet.

johrstrom: Got it, yea i think that'll work.

treydock (approved):

PR #680, opened 2022-05-16
OIDC: add OIDCResponseType to the example config

CSC-swesters:
* Explicitly add the OIDCResponseType configuration, to make users
  consider what value it should have. If the wrong configuration is
  used, OIDC authentication will fail.

**Motivation**

When testing this, we forgot to specify `OIDCResponseType` in the httpd24-httpd configurations which resulted in the wrong default value for our site. By adding the key into the configuration docs, perhaps other users will not forget to consider this.

The value in the configuration is the default value `"code"` as specified in the mod_auth_openidc upstream's comments: https://github.com/zmartzone/mod_auth_openidc/blob/master/auth_openidc.conf#L213

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202283711841049) by [Unito](https://www.unito.io)

### DIFF
--- a/source/authentication/oidc.rst
+++ b/source/authentication/oidc.rst
@@ -47,6 +47,7 @@ The following is an example :program:`ood-portal-generator` configuration file:
      OIDCPassRefreshToken: "On"
      OIDCPassClaimsAs: "environment"
      OIDCStripCookies: "mod_auth_openidc_session mod_auth_openidc_session_chunks mod_auth_openidc_session_0 mod_auth_openidc_session_1"
+     OIDCResponseType: "code"
 
 .. note::
 


source/authentication/oidc.rst
```diff
@@ -47,6 +47,7 @@ The following is an example :program:`ood-portal-generator` configuration file:
      OIDCPassRefreshToken: "On"
      OIDCPassClaimsAs: "environment"
      OIDCStripCookies: "mod_auth_openidc_session mod_auth_openidc_session_chunks mod_auth_openidc_session_0 mod_auth_openidc_session_1"
+     OIDCResponseType: "code"
```
johrstrom, line 50: If `code` is the default, shouldn't the example be something else?  I'm guessing you used `"id_token"` or `"id_token token"` (I'd also guess we use something similar).

treydock: Just FYI, OSC does not define OIDCResponseType in our configs.  WRT to tokens we set these so that things like the pre-PUN hook can bootstrap kube configs.

```
  OIDCPassClaimsAs environment
  OIDCPassIDTokenAs serialized
  OIDCPassRefreshToken On
```

CSC-swesters: Good question Jeff. Our site actually needed to set the value to `"'code id_token'"` (note the double quoting, httpd needs it to be quoted after the ood-portal-generator is run). I was debating whether to put in our example, or use the default myself before submitting the PR, but ultimately decided to use the default, since it's probably a default for a reason? I.e. admins perhaps need to modify the example configuration less, if the values are "as default as possible" so to speak.

If treydock then says that all sites don't require it, it's up for debate whether this should be added as an optional note or something instead? At least on our site, we saw errors until we explicitly specified an `OIDCResponseType`. Luckily then error message was very clear, so we found the reason instantly.

johrstrom: Cool. Well it's just an example, there's no need to get hung up on it. Thanks for the contribution!

PR #691, opened 2022-07-15
rm docs on PAM and leave it to the reader

### LINKED ISSUE #682: Remove PAM authentication
@johrstrom I wonder if we should remove the PAM authentication section? It relies on `AuthType Basic` which is incredibly insecure, for example it stores the user's password in their browser only base64 encoded so if someone hacks a personal computer, they could easily pull out that user's password.  The various PAM modules you might use like SSSD can all be done other ways like LDAP or AD through Dex or Keycloak.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202309051006096) by [Unito](https://www.unito.io)

johrstrom:
Fixes #682 

https://osc.github.io/ood-documentation-test/rm-pam-auth

Remove docs around PAM and instead have the policy of "discouraged by the authors, and left to the reader".


source/authentication/insecure.rst
```diff
@@ -0,0 +1,12 @@
+.. _authentication-insecure:
+
+Other Insecure Options
+======================
+
+There are other insecure options that you may find apache supports.
```
treydock:
```suggestion
There are other insecure options that you may find Apache supports.
```

source/authentication/insecure.rst
```diff
@@ -0,0 +1,12 @@
...
+
+It's left to the reader to try those out. Open OnDemand developers
+highly discourge users attempting to authenticate with apache's
```
treydock:
```suggestion
highly discourage users attempting to authenticate with Apache's
```

treydock (approved):

PR #696, opened 2022-07-18
use existing pages for form and submit references

johrstrom:
https://osc.github.io/ood-documentation-test/reuse-submit-yml/

Instead of having new pages for `submit.yml.erb` we should just use these pages. This also adds a config reference for `form.yml.erb` as it would likely benefit from a reference link too.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202701701476395) by [Unito](https://www.unito.io)


gerald-byrket (approved):
@johrstrom 

This works form me as well.

-gerald

PR #703, opened 2022-07-20
useradd missing username argument

simonLeary42:
previously useradd just outputs --help because there's no username specified


johrstrom:
Thanks!

PR #708, opened 2022-07-25
Add warning about using partition as user-defined attribute in app forms

### LINKED ISSUE #707: Warn about using "partition" on custom user defined attributes in the User Form
For SLURM users it might make sense to define `partition`  as a session context attribute, but this will lead to the variable set not properly passed to the script due to `partition` being a method of the Ruby mixin [enumerable](https://ruby-doc.org/core-2.6.1/Enumerable.html#method-i-partition) 

Adding a warning on that section might help other users. 

Against which branch should a PR be submitted in this case? The section `source/app-development/interactive/form.rst` is not present on the `development` branch?

i-mtz:
https://osc.github.io/ood-documentation-test/develop/how-tos/app-development/interactive/form.html
**Add your description here**

This is an attempt to fix #707 

Maybe the explanation is not totally accurate, I was not sure about how much detail to put in the warning.

Even if there is a section that already mentions how to [define partitions](https://osc.github.io/ood-documentation-test/develop/tutorials/tutorials-interactive-apps/add-custom-queue.html) having the explicit warning might help some other users. Would it make sense to re-issue the warning in the [that](https://osc.github.io/ood-documentation-test/develop/tutorials/tutorials-interactive-apps/add-custom-queue.html#add-custom-queues-partitions) section too ?

### DIFF
--- a/source/how-tos/app-development/interactive/form.rst
+++ b/source/how-tos/app-development/interactive/form.rst
@@ -474,6 +474,13 @@ The following configuration file
 
 defines a session context attribute called ``my_module_version``.
 
+.. warning::
+
+   Do not use ``partition`` as attribute name within the form. The (Ruby)
+   object created by the form is of type enumerable and ``partition`` is
+   method of that particular class. This causes the value set in the form not
+   being correctly passed to the submit script.
+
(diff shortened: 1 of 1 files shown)


johrstrom:
Thanks!

PR #721, opened 2022-08-30
remove build from source directions.

johrstrom:
https://osc.github.io/ood-documentation-test/rm-src-install/

Installing from the source code continues to be a confusion for people installing. (a) It currently reads as if it's a part of the official install process, which it isn't. People have a tendency to follow these steps blindly not realizing it's either one or the other, not both. (b) It's half documented and (c) even if they can get through a full install, such an install would be flaky.  So let's just remove it in latest as we did in develop.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202895804416630) by [Unito](https://www.unito.io)


Oglopf (approved):
Built for me and I couldn't agree more on removing this.

PR #724, opened 2022-09-06
add an example yaml for access auth

johrstrom:
https://osc.github.io/ood-documentation-test/access-oidc/

This adds a simple example for OIDC access users.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202935720683347) by [Unito](https://www.unito.io)

### DIFF
--- a/source/authentication/nsf-access.rst
+++ b/source/authentication/nsf-access.rst
@@ -18,6 +18,29 @@ Since `ACCESS`_ uses Open ID Connect (OIDC) you can see our :ref:`oidc documenta
 for more details on how to configure Open OnDemand with what CI Logon has provided in
 registering your application.
 
+
+Here's an example you can use to get started. Note that ``oidc_client_id`` and ``oidc_client_secret``
+are commented out because the are specific to your site.
+
+.. code-block:: yaml
+  :emphasize-lines: 3-4
+
+  oidc_uri: "/oidc"
+  oidc_provider_metadata_url: "https://cilogon.org/.well-known/openid-configuration"
+  # oidc_client_id: "cilogon:/client_id/..."
+  # oidc_client_secret: "..."
+  oidc_remote_user_claim: "eppn"
+  oidc_scope: "openid email org.cilogon.userinfo"
+  oidc_session_inactivity_timeout: 28800
+  oidc_session_max_duration: 28800
+  oidc_state_max_number_of_cookies: "10 true"
+  oidc_settings:
+    OIDCPassIDTokenAs: "serialized"
+    OIDCPassRefreshToken: "On"
+    OIDCPassClaimsAs: "environment"
+    OIDCStripCookies: "mod_auth_openidc_session mod_auth_openidc_session_chunks mod_auth_openidc_session_0 mod_auth_openidc_session_1"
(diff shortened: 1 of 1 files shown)


Oglopf (commented):
I'm getting a build error on the branch with the following output:

```
sphinx-build -b html -d build/doctrees  -W source build/html
Running Sphinx v1.8.0

Extension error:
Could not import extension sphinx_tabs.tabs (exception: No module named 'sphinx_tabs')
make: *** [Makefile:53: html] Error 2
```

Building on `master` and `develop` does work, so it seems to be something on this branch.

johrstrom:
> I'm getting a build error on the branch with the following output:
>
> sphinx-build -b html -d build/doctrees  -W source build/html
> Running Sphinx v1.8.0

I can tell by the version of sphinx that's not right. We're on 2.x.  I (and the CI workflows) use the container to build. It looks like you've pip installed sphinx or a virtual environment?

here's the container & command
https://github.com/OSC/ood-documentation/blob/27e31bf1eb4a73cff78781c06c1a4b2d14822da4/.github/workflows/publish.yml#L18

here's a requirements txt if you want to install a virtual env.
https://github.com/OSC/ood-documentation-build/blob/master/requirements.txt

That said - you bring up this issue: building the docs isn't super intuitive or at least when there's N options they all need to be in sync.

Oglopf:
I see the issue. I'm using the `rake docker:build` command which uses:
```
namespace :docker do

  desc "Build docs using docker"
  task :build do
    exec 'docker run --rm -i -t -v "${PWD}:/doc" -u "$(id -u):$(id -g)" ohiosupercomputer/docker-sphinx make html'
  end
end
```

And so it's that `ohiosupercomputer/docker-sphinx` image which doesn't work correctly and getting run each time I issue `rake docker:build`. 

I've never had this error before which is what seems odd. Should I open an issue?

Using the `ohiosupercomputer/ood-doc-build:v2.0.0` image builds as expected.

source/authentication/nsf-access.rst
```diff
@@ -18,6 +18,45 @@ Since `ACCESS`_ uses Open ID Connect (OIDC) you can see our :ref:`oidc documenta
 for more details on how to configure Open OnDemand with what CI Logon has provided in
 registering your application.
 
+Here's an example you can use to get started. Note that ``oidc_client_id`` and ``oidc_client_secret``
+are commented out because the are specific to your site.
```
Oglopf: `the are specific to your site` to `they are specific to your site.`

Oglopf:
Also, I opened issue #725 to address the building discussion from earlier.

source/authentication/nsf-access.rst
johrstrom:
```suggestion
are commented out because they are specific to your site.
```

johrstrom:
>  Should I open an issue?

My feeling is nearly always yes. I feel like a ticket is cheap to make and very useful long term as we're likely to forget any given oddity along the way.

source/authentication/nsf-access.rst
johrstrom: Thanks! Do you know how to make suggestions like the one I did above? They make it easy to just commit the change directly.

Oglopf (approved):
Looks good!

PR #728, opened 2022-09-21
rm gridmaps because thats not correct

johrstrom:
https://osc.github.io/ood-documentation-test/fix-access-mapping/

Gridmapping is not what we want here because that library is very out of date. And since we don't know what guidance to give, well we'll just say that while we sort out the best practice here.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203024536543139) by [Unito](https://www.unito.io)


Oglopf (approved):
LGTM.

PR #738, opened 2022-11-16
add docs on mpm configurations

johrstrom:
https://osc.github.io/ood-documentation-test/mpm-docs/

From the many discourse topics on how to improve performance - here are suggested mpm configurations. I pulled these from our production servers.  I did change `ServerLimit` and `ThreadsPerChild` and will leave a note for the same.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203387834803181) by [Unito](https://www.unito.io)

### DIFF
--- a/source/how-tos/debug/debug-apache.rst
+++ b/source/how-tos/debug/debug-apache.rst
@@ -83,35 +83,45 @@ Performance Tuning
 ------------------
 
 If you're servicing many clients at time (more than 50) you will likely need to change the
-`Apache Httpd's worker configuration`_. The default configuration may degrade service when
+`Apache Httpd's MPM configuration`_. The default configuration may degrade service when
 Httpd has to serve many clients (I.e., when you have a lot of customers using Open OnDemand).
 
 We suggest configurations similar to this. 
 
-.. warning:: 
-  `ServerLimit` and `ThreadsPerChild` should not exceed the number of cores available. 8
-  is given in the example, but will likely need to change.
+.. note:: 
+  The most important directives used to control this MPM are `ThreadsPerChild`, which controls 
+  the number of threads deployed by each child process and `MaxRequestWorkers`, which controls
+  the maximum total number of threads that may be launched.
+  
+  MaxRequestWorkers * ThreadsPerChild will be your effictive cap on the number of 
+  simultaneous requests that will be served.
+   
 
 .. code-block:: apache
 
-  # /opt/rh/httpd24/root/etc/httpd/conf.d/mpm.conf
-
+  # conf.modules.d will vary depending on the platform and version.
+  # $APACHE_HOME/conf.modules.d/mpm.conf
+  
   LoadModule mpm_event_module modules/mod_mpm_event.so
 
   <IfModule mpm_event_module>
-    ServerLimit            8
+
+    # ServerLimit is MaxRequestWorkers / ThreadsPerChild then doubled
+    ServerLimit            32
     StartServers           2
     MaxRequestWorkers      512
     MinSpareThreads        25
     MaxSpareThreads        75
-    ThreadsPerChild        8
+
+    # ThreadsPerChild value is MaxRequestWorkers / # cpus for production
+    ThreadsPerChild        32
     MaxRequestsPerChild    0
     ThreadLimit            512
     ListenBacklog          511
   </IfModule>
 
 
-.. _Apache Httpd's worker configuration: https://httpd.apache.org/docs/2.4/mod/worker.html
+.. _Apache Httpd's MPM configuration: https://httpd.apache.org/docs/2.4/mod/mpm_common.html
 .. _Apache Httpd's documentation: https://httpd.apache.org/docs/current/getting-started.html
 .. _Apache's ServerName configuration: https://httpd.apache.org/docs/2.4/mod/core.html#servername
 .. _VirtualHost: https://httpd.apache.org/docs/2.4/vhosts/
--- a/source/how-tos/debug/debug-apache.rst
+++ b/source/how-tos/debug/debug-apache.rst
@@ -88,7 +88,7 @@ Httpd has to serve many clients (I.e., when you have a lot of customers using Op
 
 We suggest configurations similar to this. 
 
-.. note:: 
+.. note::
   The most important directives used to control this MPM are `ThreadsPerChild`, which controls 
   the number of threads deployed by each child process and `MaxRequestWorkers`, which controls
   the maximum total number of threads that may be launched.
@@ -102,6 +102,7 @@ We suggest configurations similar to this.
   # conf.modules.d will vary depending on the platform and version.
   # $APACHE_HOME/conf.modules.d/mpm.conf
   
+  # MPM event is actually important for idle VNC connections alive.
   LoadModule mpm_event_module modules/mod_mpm_event.so
 
   <IfModule mpm_event_module>
--- a/source/how-tos/debug/debug-apache.rst
+++ b/source/how-tos/debug/debug-apache.rst
@@ -89,17 +89,15 @@ Httpd has to serve many clients (I.e., when you have a lot of customers using Op
 We suggest configurations similar to this. 
 
 .. note::
-  The most important directives used to control this MPM are `ThreadsPerChild`, which controls 
-  the number of threads deployed by each child process and `MaxRequestWorkers`, which controls
-  the maximum total number of threads that may be launched.
+  The most important directives are `MaxRequestWorkers`, which controls the number of simultaneous
+  requests and `ServerLimit` combined with `ThreadsPerChild`, which when multiplied must be at least as big as
+  the value for `MaxRequestWorkers`.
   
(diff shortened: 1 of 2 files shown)


source/how-tos/debug/debug-apache.rst
```diff
@@ -77,7 +79,39 @@ Or you're using the wrong hostname in your browser.
...
+  is given in the example, but will likely need to change.
+
+.. code-block:: apache
+
+  # /opt/rh/httpd24/root/etc/httpd/conf.d/mpm.conf
+
+  LoadModule mpm_event_module modules/mod_mpm_event.so
+
+  <IfModule mpm_event_module>
+    ServerLimit            8
```
johrstrom, line 116: `ServerLimit` and `ThreadsPerChild` should not exceed the number of cores available so I left this to 8 given that may just fit a lot of sites (we have 32).

source/how-tos/debug/debug-apache.rst
```diff
@@ -77,7 +79,39 @@ Or you're using the wrong hostname in your browser.
...
+
+We suggest configurations similar to this. 
+
+.. warning:: 
+  `ServerLimit` and `ThreadsPerChild` should not exceed the number of cores available. 8
+  is given in the example, but will likely need to change.
+
+.. code-block:: apache
+
+  # /opt/rh/httpd24/root/etc/httpd/conf.d/mpm.conf
```
treydock: This is only the path on RHEL7 and this path is specific to Puppet.  The path on default install is `/opt/rh/httpd24/root/etc/httpd/conf.modules.d/00-mpm.conf`

source/how-tos/debug/debug-apache.rst
```diff
@@ -77,7 +79,39 @@ Or you're using the wrong hostname in your browser.
...
+
+  LoadModule mpm_event_module modules/mod_mpm_event.so
+
+  <IfModule mpm_event_module>
+    ServerLimit            8
+    StartServers           2
+    MaxRequestWorkers      512
+    MinSpareThreads        25
+    MaxSpareThreads        75
+    ThreadsPerChild        8
```
treydock: Per comment in Puppet:

```
# This value is maxrequestworkers / # cpus for production
```

Just how I determined the value

source/how-tos/debug/debug-apache.rst
treydock: Per comment in Puppet:

```
# This value is maxrequestworkers / threadsperchild then doubled
```

source/how-tos/debug/debug-apache.rst
treydock: Also don't put module configs in `conf.d` that's for VirtualHost configs, module configs in modern Apache got in `conf.modules.d` at least on RHEL hosts.

source/how-tos/debug/debug-apache.rst
```diff
@@ -77,7 +79,39 @@ Or you're using the wrong hostname in your browser.
...
+------------------
+
+If you're servicing many clients at time (more than 50) you will likely need to change the
+`Apache Httpd's worker configuration`_. The default configuration may degrade service when
+Httpd has to serve many clients (I.e., when you have a lot of customers using Open OnDemand).
+
+We suggest configurations similar to this. 
+
+.. warning:: 
+  `ServerLimit` and `ThreadsPerChild` should not exceed the number of cores available. 8
```
treydock: Per other comment, threads per child can very easily exceed number of cores as you typically can have a lot more threads on a host than number of CPUs.

source/how-tos/debug/debug-apache.rst
johrstrom: Thanks Ive updated all of these.

source/how-tos/debug/debug-apache.rst
```diff
@@ -77,7 +79,50 @@ Or you're using the wrong hostname in your browser.
...
+
+We suggest configurations similar to this. 
+
+.. note::
+  The most important directives used to control this MPM are `ThreadsPerChild`, which controls 
+  the number of threads deployed by each child process and `MaxRequestWorkers`, which controls
+  the maximum total number of threads that may be launched.
+  
+  MaxRequestWorkers * ThreadsPerChild will be your effictive cap on the number of 
+  simultaneous requests that will be served.
```
johrstrom: Pulled the first bit from apache docs itself. I believe the 2nd line is true. That seems like the starting point. This _could_ be sufficient, but I feel like we need a table or something as a decent starting point? 

Or just say `512` is likely to work well on most systems? and if you have a particularly small system (less than what, 2 cores 4 GB?) maybe `256`?

treydock: The number I came up with was a combination of cores on the system and max expected concurrent traffic. Once the worker threads are exhausted, people won't be able to access OnDemand so it's important to keep the number well above the expected max usage.

johrstrom: Well most folks are going to copy paste what we have I'm sure, so I guess I'm asking if 512 will work on most systems if most systems have more than 8 cores (which I have no idea if that's true, most could have 4).

source/how-tos/debug/debug-apache.rst
```diff
@@ -77,7 +79,50 @@ Or you're using the wrong hostname in your browser.
...
+   
+
+.. code-block:: apache
+
+  # conf.modules.d will vary depending on the platform and version.
+  # $APACHE_HOME/conf.modules.d/mpm.conf
+  
+  # MPM event is actually important for idle VNC connections alive.
+  LoadModule mpm_event_module modules/mod_mpm_event.so
```
johrstrom: This file still has the `LoadModule` statement which I'm sure is not the best practice, but IDK how else to convey that you also need to do this (and different platforms have different schemes).

source/how-tos/debug/debug-apache.rst
treydock: It's better to have fewer ServerLimit and more threads, the server limit is number of processes I believe so that more tied to cores where as threads are less dependent on number of cores.

source/how-tos/debug/debug-apache.rst
treydock: `LoadModule` is necessary to load this MPM but it's not always in same file as the config but doesn't hurt to have them in same file.  It's also important people remove referenecs to `mod_mpm_prefork.so` which is the default.  So by default there is a line like this:

```
LoadModule mpm_prefork_module modules/mod_mpm_prefork.so
```
That line must be deleted and replaced by one to load event MPM.  Can't have both loaded at same time, pretty sure Apache will refuse to start if you do.

johrstrom: Updated the comment, it appears to be commented out in a default install.

source/how-tos/debug/debug-apache.rst
johrstrom: OK - I've changed all these values. Given that most folks are going to copy paste & even fewer will update after the fact I'd like to get a sizing that would likely work OK for most sites. So I assumed 8 cores, which actually could be high.

source/how-tos/debug/debug-apache.rst
```diff
@@ -77,7 +79,57 @@ Or you're using the wrong hostname in your browser.
...
+
+.. note::
+  The most important directives used to control this MPM are `ThreadsPerChild`, which controls 
+  the number of threads deployed by each child process and `MaxRequestWorkers`, which controls
+  the maximum total number of threads that may be launched.
+  
+  MaxRequestWorkers * ThreadsPerChild will be your effictive cap on the number of 
+  simultaneous requests that will be served.
+
+  The example configuration below can handle about 8000 simultaneous requests.
```
treydock: It can only handle 256 requests at the same time, that's what `MaxRequestWorkers` defines.

treydock: I believe it's also set by `ThreadLimit`

johrstrom: >  MaxRequestWorkers * ThreadsPerChild will be your effictive cap on the number of 
  simultaneous requests that will be served.

I take it this isn't right?  I read that as you have workers that spawn children, so every child can handle a request. The number of children being `MaxRequestWorkers * ThreadsPerChild`.

treydock: https://httpd.apache.org/docs/2.4/mod/mpm_common.html#maxrequestworkers

```
The MaxRequestWorkers directive sets the limit on the number of simultaneous requests that will be served.
```

ThreadsPerChild just describes the distribution of threads across the actual processes spawned.  With event MPM each request will be a thread where as prefork (default) each request is a unique process.

johrstrom: Can you take this PR over? I feel like you've got more sensibility to take it forward.

Really I just want some config that'll work for 50% or more of the folks who copy & paste it.

treydock: I made my edits.

treydock (approved):

PR #754, opened 2022-12-23
UserConfiguration Properties + Profiles documentation

abujeda:
Updates to the ondemand.d/*.yml files page , adding the new UserConfiguration properties that can now be configured using these files.

As well, added basic information about profiles and how they work.

Added placeholders for the new documentation for:

Custom navbar
Custom help bar
Custom interactive apps menu
Custom pages

This is to be able to create a reference link from this page.

This PR is work in progress, I just wanted to start with the documentation for the new features that we have integrated and get early feedback on how to approach the documentation changes needed.

This is the page that this PR modifies:
https://osc.github.io/ood-documentation/latest/reference/files/ondemand-d-ymls.html

PR #760, opened 2023-01-16
Havard iqss collaboration features

abujeda:
PR with all the doc changes from the features integrated from the collaboration with Harvard IQSS.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203757444101754) by [Unito](https://www.unito.io)


source/index.rst
```diff
@@ -65,6 +65,7 @@ These are institutions who were early adopters or provided HPC resources for dev
   how-tos/debug
   enable-desktops
   how-tos/app-development
+  how-tos/support_ticket
```
johrstrom: I don't think this should be a top level menu item.

abujeda: OK. Will move into one of the existing sections

source/how-tos/support_ticket.rst
```diff
@@ -0,0 +1,132 @@
...
+        subject:
+          value: "OnDemand Problem: "
+      form:
+        - email
+        - subject
+        - department
+
+  .. code-block::
+
+    # /etc/ood/config/apps/dashboard/views/support_ticket/email/custom_template.txt.erb
```
johrstrom:
```suggestion
    # /etc/ood/config/apps/dashboard/views/support_ticket/email/_custom_template.txt.erb
```

Is this right?

source/how-tos/support_ticket.rst
```diff
@@ -0,0 +1,132 @@
...
+
+.. note::
+  
+  Please note that an ``email`` field must be present in the form for the support ticket feature to be funtional.
+
+Custom fields can be added, but they will require a custom email template to make use of the field values to generate the email body.
+Drop the template file into ``/etc/ood/config/apps/dashboard/views/support_ticket/email`` and reference it
+in the configuration. This template file can be any format Rails recognizes for emails, usually the ``.txt.erb`` extension.
+
+The example below shows a custom template configuration with 3 fields and how to use them in the template to generate the email body.
```
johrstrom, line 98: Can you also mention here that this is the same structure as an interactive app's form?

source/how-tos/support_ticket.rst
```diff
@@ -0,0 +1,132 @@
...
+.. figure:: /images/support_ticket_custom_form.png
+   :align: center
+
+Troubleshooting
+...............
+
+.. tip::
+
+  Ensure that your Dashboard Rails environment is configured properly to send emails and to raise errors.
+  Update your environment settings accordingly ``dashboard/config/environments/(development.rb|test.rb|production.rb)``
```
johrstrom: What is this? Real sites should not be modifying these files.

source/how-tos/support_ticket.rst
```diff
@@ -0,0 +1,132 @@
+.. _support_ticket_guide:
+
+Support Ticket
```
johrstrom, line 3: This section doesn't have anything on `rt` support. Can you add that as well?

abujeda: Will do.

source/how-tos/support_ticket.rst
abujeda: I mentioned that form updates follow the same structure as interactive apps at the beginning of the `Customizations` section. Shall I repeat it here as well?

source/how-tos/support_ticket.rst
abujeda: That is correct. For email templates, Rails does not require the `_`.
I tested the configuration while doing the documentation.

source/how-tos/support_ticket.rst
johrstrom: nope, sorry, disregard! I misread that bit.

source/how-tos/support_ticket.rst
johrstrom: Oh because we pull it, not because it's a part of the rails infrastructure. 

Hmmmmm, this is different than all the other partials they can overload then.  I wonder if we should force the prepended `_` so that it's like every other partial.

source/how-tos/support_ticket.rst
abujeda: This is only needed if there are problems with the integration and requires debugging in a local environment.
In the `development` environment, Rails `mailer` will not raise any errors, so if admins want to debug issues locally, they need to set this settings.

I added this section as it helped me debug issues while developing the feature.
Shall I remove it?

johrstrom: I see. I don't know if any folks use the development environment. Usually they just deploy on a test system and interact with the production system.

> In the development environment, Rails mailer will not raise any errors, 

We should just reconfigure `development.rb` upstream so that it does.

> Shall I remove it?

Yes please.

source/how-tos/support_ticket.rst
abujeda: I think we can use partials with email templates too. We could create en email layout template and include the default or custom as partials. This will follow the same structure as the rest of the template overrides in OnDemand.

In this case, the email layout template will be empty, just with the render partial statement.

source/how-tos/support_ticket.rst
abujeda: OK. Will remove this section and add the changes to the `development.rb` to raise errors for emails.

source/how-tos/support_ticket.rst
abujeda: Yes we can use partials. I will update to use partials and update the docs.
This will make consistent with everything else

source/index.rst
abujeda: Moved under `Configuration Reference`

source/how-tos/support_ticket.rst
abujeda: Completed

source/how-tos/support_ticket.rst
abujeda: removed

johrstrom:
I'm ready for this if you are. Do you want to squash? Also I assume this supercedes #754 so I'm guessing we close that one?

abujeda:
Ready 🚀 
Squash completed. Yes, this supersedes #754 (will close it)

abujeda:
Will continue reviewing the docs and will make PRs if changes are needed.

Thanks

PR #771, opened 2023-01-30
update ACCESS IdP docs

jbasney:
https://osc.github.io/ood-documentation-test/latest/authentication/nsf-access.html

* Link directly to the ACCESS IdP docs rather than the ACCESS IAM landing page.
* Reference CILogon in the OIDC section, not the SAML section.
* Update OIDC example configuration to use recommended oidc_remote_user_claim and oidc_scope values.
* SAML metadata is now published by InCommon.
* Update link for access-oauth-mapfile.


johrstrom (approved):
Thanks!

PR #778, opened 2023-02-08
2.1 Feature documentation improvements after review

abujeda:
Improvements to the docs for the new 2.1 features after a review.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203931335296149) by [Unito](https://www.unito.io)


source/customizations/profiles.inc
```diff
@@ -23,13 +23,12 @@ logic will aggregate all profiles into a single consolidated object.
       config_property: config_value
 
 OnDemand uses an inheritance and override approach for profiles.
-The root configuration provides default properties that all profiles inherit.
-Then, profile properties can override these values or define new ones as needed.
+The root profile, this is the default profile created by the system based on all the properties defined outside any profile definition,
+provide the default values that all profiles inherit. Then, profile properties can override these values or define new ones as required.
```
johrstrom: I almost liked the first sentence better. Mabye break this up into 2 sentences.

Maybe something like this?
```suggestion
The root profile is the default profile created by the system based on all the properties defined outside any profile definition.  All other profiles will inherit these values.

Then, profile properties can override these values or define new ones as required.
```

abujeda: Fixed

PR #792, opened 2023-03-08
correct the systemd file used for ood-portal

mrgum:
the systemd file specified was the ood one not the ood-portal one where these options reside

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204136607995597) by [Unito](https://www.unito.io)

### DIFF
--- a/source/reference/commands/ood-portal-generator.rst
+++ b/source/reference/commands/ood-portal-generator.rst
@@ -97,8 +97,8 @@ Options
    .. warning::
       The systemd file for Apache will run the update_ood_portal script with defaults
       and will not use a different file, rendering this option obsolete unless you also
-      modify systemd config (``/etc/systemd/system/httpd24-httpd.service.d/ood.conf`` in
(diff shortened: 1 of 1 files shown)


johrstrom:
Hi, thank you for the contribution. I went to check on this and apparently we create both.  

```
[root@68d97ebf4c72 ~]# ls /etc/systemd/system/httpd.service.d/ -lrta
total 16
-r--r--r-- 1 root root   63 Feb 13 16:34 ood.conf
drwxr-xr-x 1 root root 4096 Feb 13 18:03 ..
-r--r--r-- 1 root root  242 Feb 13 18:03 ood-portal.conf
drwxr-xr-x 1 root root 4096 Feb 13 18:03 .
```

I'm going to have to check in the source as to why this is the case.

johrstrom:
Thanks for the contribution! You're right that these docs should point to `ood-portal.conf`.

PR #815, opened 2023-05-18
Add conn-params link and page with content

Oglopf:

https://osc.github.io/ood-documentation-test/document-conn-params/

**Add your description here**
Documenting the conn-params pattern.

This still needs a bit of work as I was surprised when working the Rstudio example to notice not all variables were exportet, which I had thought was necessary.

Not a big deal, but I'd like to understand when the `export` is needed and not before this is done.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204629982865110) by [Unito](https://www.unito.io)

### DIFF
--- a/source/how-tos/app-development/interactive/conn-params.rst
+++ b/source/how-tos/app-development/interactive/conn-params.rst
@@ -36,7 +36,7 @@ The files which can be adjusted to::
   └── view.html.erb
 
 And the ``submit.yml.erb`` will use ``conn_params`` to set the custom variables to pass back to the ``view`` to 
-be rendered, but also it seems that is not true, Rstudio does not do so for ``port`` or ``host``. ....hmmmm::
+be rendered::
 
     ---
     batch_connect:
--- a/source/how-tos/app-development/interactive/conn-params.rst
+++ b/source/how-tos/app-development/interactive/conn-params.rst
@@ -1,14 +1,18 @@
 .. _app-development-interactive-conn-params:
 
-Connection Parameters (connections.yml)
-=======================================
+Connection Parameters ``conn_params``
+=====================================
 
-The `connections.yml` configuration file facilitates communication between 
-the backend server and an app, via the ``before.sh.erb``, ``submit.sh.erb``, 
-and ``view.html.erb`` files.
+App developers can use ``conn_params`` in the ``submit.yml.erb`` to pass runtime data generated
+in the ``before.sh.erb`` back to the ``view.html.erb``. 
 
-This configuration can be crucial when specific data, such as a ``csrf_token``, 
-is required for operations like logging into RStudio.
+This is helpful for:
+
+* Data that is only known **after** the job submits and starts running.
+* Data that needs to be used to connect to the application.
+
+This technique will generate a file in the app's root called ``connection.yml`` when the app launches 
+which will contain the defined variables and their associated values.
 
 Configuration
 -------------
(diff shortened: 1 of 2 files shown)


johrstrom:
> Not a big deal, but I'd like to understand when the export is needed and not before this is done.

I'm fairly sure you you need to export everything in the `before.sh.erb` . You can look at the flow here - but `script.sh.erb` is running in the background (so it's parallel to the main job_script_content.sh that writes the yaml). so it's best to export everything in `before.sh.erb` because that's synchronous WRT to the script that writes the yaml.

https://github.com/OSC/ood_core/blob/95c4799f1d5bcf0f314e37fd8bd2d53b99edd131/lib/ood_core/batch_connect/template.rb#L305

source/how-tos/app-development/interactive/conn-params.rst
```diff
@@ -0,0 +1,108 @@
...
+
+Connection Parameters (connections.yml)
+=======================================
+
+The `connections.yml` configuration file facilitates communication between 
+the backend server and an app, via the ``before.sh.erb``, ``submit.sh.erb``, 
+and ``view.html.erb`` files.
+
+This configuration can be crucial when specific data, such as a ``csrf_token``, 
+is required for operations like logging into RStudio.
```
johrstrom: I believe it's `connection.yml` singular.

I'm trying to figure out a way to tighten this phrasing. One thing is it's a 1 way communication from `before.sh.erb -> view.html.erb` so the language should reflect that.

The other thing I think we need to stress is that these values are generated during runtime.

I think the big thing we need to communicate here in the first paragraph is _why_ anyone would need these.

Here's my first stab at what that may look like:
```
App developers can use `conn_params` in `submit.yml.erb` to pass runtime data generated
in the `before.sh.erb` back to the `view.html.erb`. 

This is helpful for 
* Data that is only known _after_ the job submits and starts running.
* Data that needs to be used to connect to the application.
```

source/how-tos/app-development/interactive/conn-params.rst
```diff
@@ -0,0 +1,108 @@
...
+
+  my_app/
+  ├── submit.html.erb
+  ├── templates/
+  │   ├── before.sh.erb
+  └── view.html.erb
+
+The primary purpose of this configuration is to utilize the ``before.sh.erb`` script 
+to generate a piece of data. This data is then passed to the ``submit.sh.erb`` which 
+then will hand off to the ``view.html.erb`` for rendering in the interactive apps page.
```
johrstrom:
```suggestion
to generate a piece of data. This data is then written to a ``connection.yml`` file which 
the ``view.html.erb`` reads for rendering the button to connect to the application.
```

source/how-tos/app-development/interactive/conn-params.rst
Oglopf: Well what confused me was that I saw things like `password` and `host` in the Rstudio app being used in the `view` that hat not been exported.

I think we should just tell people to `export` regardless. The simple fix to my request would be for me to go make the change to that code and include the `export` now that I think on it lol.

Also, my advice to anyone to clean up wording or get phrasing right is to take what's there, put it in ChatGPT, and explain what you want. It is a very good editor to work through wording with (though not perfect or always right).

johrstrom: > Well what confused me was that I saw things like password and host in the Rstudio app being used in the view that hat not been exported.

Now that you mention it, I suspect it's because we `source` the `before.sh.erb`. I think when you `source` things you don't have to `export`?

I see them being defined, just not exported.

In any case - I think my comment is more around making sure we explain the use case before details of implementation.

source/how-tos/app-development/interactive/conn-params.rst
```diff
@@ -0,0 +1,94 @@
...
+  my_app/
+  ├── submit.html.erb
+  ├── templates/
+  │   ├── before.sh.erb
+  │   └── dir
+  │       └── another_script.sh.erb
+  └── view.html.erb
+
+And the ``submit.yml.erb`` will use ``conn_params`` to set the custom variables to pass back to the ``view`` to 
+be rendered, but also it seems that is not true, Rstudio does not do so for ``port`` or ``host``. ....hmmmm::
```
johrstrom: I think this is a typo - `....hmmmm::`

Oglopf:
```suggestion
be rendered::
```

johrstrom (approved):

PR #855, opened 2023-09-04
Improvements to required fields for the support ticket configuration

abujeda:
Part of the fixes for #2867

Improvements to the support ticket documentation required fields


johrstrom:
Can you branch off of `latest` instead? We updated our doc workflow so that we can commit kinda freely to `latest` (the current documentation branch) and there's CI to take that update and forward port it to the `develop` (the next version's documentation branch).

It's just a little trick so we can keep the `latest`/current documentation as up to date with fixes like this.

PR #881, opened 2023-10-27
Add login cluster only recipe

Oglopf:

https://osc.github.io/ood-documentation-test/login-cluster-only-recipe/

**Add your description here**
Adding section to cluster configuration to show users how to create a cluster config that is only used for logins and not to run jobs.

### DIFF
--- a/source/installation/cluster-config-schema.rst
+++ b/source/installation/cluster-config-schema.rst
@@ -190,3 +190,28 @@ See :ref:`Disable Host Link in Batch Connect Session Card <disable-host-link-bat
 .. note::
 
   The user is responsible for providing the `%s` that is used to place the script content. If a `script_wrapper` is provided without `%s` then batch connect applications are unlikely to work properly.
+
+
+******************
+Login Cluster Only
+******************
+
+Suppose you need want to create a *login cluster that does not schedule or run jobs*. It is used purely as a login/shell cluster only.
+
+To accomplish this, you need to simply leave out the ``v2job`` stanza that associates a scheduler with the cluster.
+
+An example config file in ``ondemand.d/pitzer_01_login.yml``:
+
+.. code-block:: yaml
+
+    ---
+    v2:
+    metadata:
+        title: "Pitzer Login"
+        url: "https://www.osc.edu/supercomputing/computing/pitzer"
+        hidden: false
+    login:
+        host: "pitzer-login01.hpc.osu.edu"
+
+Again, the thing to not here is we've left off the ``v2.job`` which renders the cluster useable only for logins, i.e.
+*no jobs will be scheduleable on this cluster.*
--- a/source/installation/cluster-config-schema.rst
+++ b/source/installation/cluster-config-schema.rst
@@ -213,5 +213,5 @@ An example config file in ``ondemand.d/pitzer_01_login.yml``:
     login:
         host: "pitzer-login01.hpc.osu.edu"
 
-Again, the thing to not here is we've left off the ``v2.job`` which renders the cluster useable only for logins, i.e.
+Again, the thing to note here is we've left off the ``v2.job`` which renders the cluster useable only for logins, i.e.
 *no jobs will be scheduleable on this cluster.*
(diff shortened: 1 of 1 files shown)


source/installation/cluster-config-schema.rst
```diff
@@ -190,3 +190,28 @@ See :ref:`Disable Host Link in Batch Connect Session Card <disable-host-link-bat
...
+    ---
+    v2:
+    metadata:
+        title: "Pitzer Login"
+        url: "https://www.osc.edu/supercomputing/computing/pitzer"
+        hidden: false
+    login:
+        host: "pitzer-login01.hpc.osu.edu"
+
+Again, the thing to not here is we've left off the ``v2.job`` which renders the cluster useable only for logins, i.e.
```
johrstrom:
```suggestion
Again, the thing to note here is we've left off the ``v2.job`` which renders the cluster useable only for logins, i.e.
```

source/installation/cluster-config-schema.rst
```diff
@@ -190,3 +190,28 @@ See :ref:`Disable Host Link in Batch Connect Session Card <disable-host-link-bat
...
   The user is responsible for providing the `%s` that is used to place the script content. If a `script_wrapper` is provided without `%s` then batch connect applications are unlikely to work properly.
+
+
+******************
+Login Cluster Only
+******************
+
+Suppose you need want to create a *login cluster that does not schedule or run jobs*. It is used purely as a login/shell cluster only.
+
+To accomplish this, you need to simply leave out the ``v2job`` stanza that associates a scheduler with the cluster.
```
johrstrom:
```suggestion
To accomplish this, you need to simply leave out the ``v2.job`` stanza that associates a scheduler with the cluster.
```

source/installation/cluster-config-schema.rst
```diff
@@ -190,3 +190,28 @@ See :ref:`Disable Host Link in Batch Connect Session Card <disable-host-link-bat
 .. note::
 
   The user is responsible for providing the `%s` that is used to place the script content. If a `script_wrapper` is provided without `%s` then batch connect applications are unlikely to work properly.
+
+
+******************
+Login Cluster Only
+******************
```
johrstrom: This renders fine, but I feel like it's weird to have the `****` line on the top  - do headers in other files do this?

Oglopf: I'm not sure what the comment means with "`****` line on the top". I'm just trying to add the subtopic under cluster, and that was the only way I could find that matched what was in the file and others to make this visible in the nav, while also keeping this in a somewhat rational place.

Were you thinking a separate page like the resource-providers and a navigation link at that depth instead of under cluster config?

johrstrom: I mean the line of **** on 195 above the heading. 

This I've seen a lot throughout this repo.
```
Login Cluster Only
******************
```

This - with a line of **** above the heading - I don't think I've seen in this repo.
```
******************
Login Cluster Only
******************
```

Oglopf: Yeah, we use that all over. I think you are thinking of `====` and some others. You get an error without that top line.

If you look in the `cluster-config-schema.rst` for instance, you can see it in `A Break Down`.

I like it because it seemed to give the best navigation output for a user to see this is all, but I am open to changing it.

Oglopf: Huh, but yeah when I look in `kubernetes.rst` I only see the version with them on the bottom. Well then, I am happy to remove the top line, less text the better.

Oglopf: Yeah I don't know, it's a build error around the title being at an inconsistent level, but when you add that top line back things work. 

This is a small thing to me that it needs that though, and I see it used elsewhere so it doesn't bother me.

Oglopf: Haha, so I think the reason this file needs that is because the rest of the file was written that way.

When I remove the top lines on the other headings for `First an Example` and `A Break Down` and have only bottom lines for them all with `****` it will render normal and show correctly in the nav. So i'll include a commit to remove those other top pieces as well.

PR #899, opened 2023-12-04
rm module enable for rhel9

### LINKED ISSUE #814: RHEL 9 dependencies could be wrong
There seems to be an issue with how RHEL/9 is documented. See this discourse topic.

https://discourse.openondemand.org/t/ood3-installation-on-rhel9/2692/2

I'm finding that our CI that builds our packages runs ruby 3.0 which is fine, but nodejs version is actually 16. At least in `EL 9.1`. 

From the discourse topic, it could be the case that these packages have been updated in other minor versions of EL 9 and/or streams changed.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1204613972677264) by [Unito](https://www.unito.io)

johrstrom:
https://osc.github.io/ood-documentation-test/rhel9-deps/

Fixes #814  - though I don't know if it'll stay like this long term. We build against the default modules, so I'm guessing it'll last, but can't say that for sure. @treydock please advise.

### DIFF
--- a/source/installation/install-software.rst
+++ b/source/installation/install-software.rst
@@ -83,7 +83,6 @@ Some operating systems use `Software Collections`_ to satisfy these.
       .. code-block:: sh
 
          sudo dnf install epel-release
-         sudo dnf module enable ruby:3.0 nodejs:16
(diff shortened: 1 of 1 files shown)


johrstrom:
This is duplicate of #904.

PR #912, opened 2023-12-12
add docs to develop dashboard and shell apps within ood itself

Oglopf:

https://osc.github.io/ood-documentation-test/ood-in-ood-dev/

**Add your description here**
This PR adds sections on how to use the sandbox capability of OOD to do development work, from within the production system, on a separate dev dashboard or shell app.

This also includes tips in presenting all this in how you can set new environment variables for this dev dashboard along with how to set your very own configuration directory for the dev dashboard as well.

Similarly, the shell app entry shows a user how to make their own dev shell, how to make changes to this shell, and then how to rebuild correctly to see the changes.

I'd like to add some more to this such as work on the Files app as well, the goal being to show the world how we do this internally at OSC.


Oglopf:
# To-do
- [x] add some pictures for the dev dashboard with color variation
- [x] cleanup wording 
- [x] a couple more advanced examples around customization.
- [x] the shell app page is are bones. I'd like to enrich it a bit still and walk through a rebuild when changes are made

To be clear, the goal here is that these docs would largely be used for onboarding and community dev work. Now, if a new dev comes on board, we can point them to our docs and get them working rather than have many places with many tasks which we expect them to ingest and compose correctly in their head. But, that's just analysis for anlysis' sake. This gives the moves up front, the algebra to build is right there and hopefully this creates a much faster on ramping.

This also let's user's who've fallen into the usual trap of "I started this on a VM, showed it to a researcher, and next thing knew it was a service!" problem. Now, they can at least have a dev environment to work in rather than posting every question on the Discourse, thereby freeing us up more as well.

Oglopf:
I think I'm happy with this at this point and good with a review and pulling it in. We can enrich the entries more if desired later, but I think this at least gives good fundamentals on how to do this type of work on ood.

source/tutorials/tutorials-dashboard-apps/dashboard.rst
```diff
@@ -0,0 +1,142 @@
...
+----------------------
+#. ``clone`` the OOD repo into: ``cd ~/ondemand/misc`` then ``git clone git@github.com:OSC/ondemand.git``
+#. Work out of the ``dev`` directory/space: ``cd ~/ondemand/dev/``.
+#. Symlink to the ``dashboard`` app in the cloned repo: ``ln -s ../misc/ondemand/apps/dashboard/dashboard/ dashboard``
+#. Make a branch and to begin word on your dev dashboard: ``git checkout -b dev_work`` 
+#. Rebuild the ood dashboard app:  ``bin/setup``.
+#. Navigate to the Sandbox and launch your ``dashboard`` app to use the ``dev_work``.
+
+Notice the ``url`` for this ``dev_dashboard`` app. No longer do 
+you  see ``*/sys/dashboard/`` but instead ``*/dev/dashboard/``.
```
johrstrom: I think the `dev_dashboard` should be `dev/dashboard` no?  Plus - why replace `/pun` with `*` Can't we just say `/pun`?

```suggestion
Notice the ``url`` for this ``/pun/dev/dashboard`` app. No longer do 
you  see ``/pun/sys/dashboard/`` but instead ``/pun/dev/dashboard/``. 
```

source/tutorials/tutorials-dashboard-apps/shell-app.rst
```diff
@@ -0,0 +1,48 @@
...
+And then step into that symlinked directory and run:
+#. ``bin/setup`` 
+
+If you go start this app from sandbox you'll notice a different url than before, 
+showing you are in a dev shell: ``pun/dev/shell-1.8``
+
+Issues and Errors
+.................
+**If above fails** try to purge the modules with ``module purge`` on the login node first, 
+then ensure the *correct ruby version* with ``module load ruby/3.0.2`` then rebuild with ``bin/setup`` as above.
```
johrstrom: Is this in the right place? It's talking about `ruby` module but it's in the shell's tutorial?

I do however like the idea of having an issues section that can help either/both tutorials.

johrstrom: Also note that the ruby version is subject to change (it's 3.1 in OOD 3.1) but also dependent on the environment.  We have `ruby/3.0.2` but any ruby 3.0 version would work.

Though again, it turns into 3.1 in OOD 3.1 so we may need to use/create a jinja variable for the ruby version (there may already be one).

source/tutorials/tutorials-dashboard-apps/dashboard.rst
```diff
@@ -0,0 +1,142 @@
+.. _app-development-tutorials-dashboard-apps-dashboard:
+
+Developing the Dashboard App
+============================
+
+.. warning::
+    You muse have followed :ref:`enabling-development-mode` for any part of this tutorial to work.
+
+Strategy
+--------
```
johrstrom: Is `Strategy` meant to convey something different than the actual steps? Or is it like a summation before you get started? I'm a little confused on the purpose of this section.

That said - I did think a `Things to know` section may help (maybe that's what I thought this section was trying to convey?). Like general stuff to be aware of that doesn't necessarily fit anywhere else.

Oglopf: Because the strategy applied to the shell as well, so it seemed like a pttern to call out overall is all. But, I'd be fine removing it. I don't like Things To Know sections myself, I prefer steps like this which layout the strategy, but maybe others would?

source/tutorials/tutorials-dashboard-apps/shell-app.rst
Oglopf: I think that is a remnant of the old notes I built this from, which itself was copied largely from the dashboard notes, so that's how that snuck in I think. I can remove that.

source/tutorials/tutorials-dashboard-apps/dashboard.rst
Oglopf: I added a "Things to know" section. Let me know if that is what you had in mind, and feel free to enrich it. I tried to just call out anything that bit me starting out.

johrstrom: I guess I'm just wondering how it's different from _the actual steps_ - is it an overview or something? 

And even then the notes are confusing when looking at the steps. Take symlinking for example it says:

https://github.com/OSC/ood-documentation/blob/b3a6a71f583ef2a8ad89f12e6d166f33912d8824/source/tutorials/tutorials-dashboard-apps/dashboard.rst?plain=1#L13

Then the steps say:
https://github.com/OSC/ood-documentation/blob/b3a6a71f583ef2a8ad89f12e6d166f33912d8824/source/tutorials/tutorials-dashboard-apps/dashboard.rst?plain=1#L23

The name of the section aside - it seems to me that the document could use some complimentary information around _why you're doing any given step_.  Like we're symlinking because  `OOD will try to recognize any directory in ~/ondemand/dev as an application`.

PR #920, opened 2024-01-02
kilobyte -> kibibyte

simonLeary42:
while configuring quota warnings, I converted to "blocks" by dividing "number of bytes" by 1000, because it said `KB`. But in the source code, it's using 1024.

https://github.com/OSC/ondemand/blob/5ae612e418d5be1654d663a7edd8da29d66d9b70/apps/dashboard/app/models/quota.rb#L7


johrstrom (approved):
Thanks!

PR #926, opened 2024-01-26
Add information about ProxyTimeout

buzh:

https://osc.github.io/ood-documentation-test/BRANCH-NAME/

**Add your description here**

The websockify connection will occasionally drop when the VNC session sits idle for a while, producing messages in vnc.log like this: 

`Client 127.0.0.1 gone`

To reconnect, the user cannot press the noVNC button, but has to go back to OOD and click the card button since. In order to avoid this, a simple fix is just increasing the timeout that mod_proxy uses.

It might be worth setting this value by means of the portal config generator, but for now I just made this PR to the docs so that others don't need to spend as much time as I did chasing the cause of this issue.

### DIFF
--- a/source/customizations.rst
+++ b/source/customizations.rst
@@ -1325,3 +1325,17 @@ For completed sessions, the system will only show the delete button.
 
 .. _OSC's rclone documentation: https://www.osc.edu/resources/getting_started/howto/howto_use_rclone_to_upload_data
 .. _2.0 documentation for controling the navbar: https://osc.github.io/ood-documentation/release-2.0/customization.html#control-which-apps-appear-in-the-dashboard-navbar
+
+Preventing VNC from timing out when inactive
+--------------------------------------------
+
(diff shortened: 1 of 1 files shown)


johrstrom:
Hi - I'm sorry to do this to you but I'm going to close this. The reason is 2 fold: First, this PR prompted me to make sure that this update in ood_core got put into 3.1 https://github.com/OSC/ood_core/pull/826. This update made it to 3.1.4 (though I renamed the variable in case you want to use it), and I've been using it with some success. So with heartbeats enabled the connection can stay alive more easily without increasing the timeout here. That leads to the 2nd point I made above, I'm not entirely sure that setting such a high `ProxyTimeout` is advisable. Enabling heartbeats makes the point a bit moot, but I'd be worried about tying up so many resources in apache.

In any case, thank you so much for the contribution even if it's not getting merged. I apologize that it didn't get merged, but it did push me to make sure the heartbeats update made it to 3.1. I hope you understand and see that heartbeats solve this problem much better than increasing the timeout.

Thanks (and sorry) again!

PR #927, opened 2024-02-01
Correct copy_enviornment to copy_environment on Slurm cluster configuration page

dhml:
While browsing the [documentation](https://osc.github.io/ood-documentation/latest/installation/resource-manager/slurm.html) I noticed that the `copy_environment` option is incorrectly specified as `copy_enviornment`.  Corrected here.


johrstrom (approved):
Thanks!

PR #951, opened 2024-03-13
Add globus example for multiple endpoints

### LINKED ISSUE #948: Add Globus Doc Example
In Open Office hours today we realized we should add an example or small gif to demonstrate how to use the globus button to show how the fs and endpoints work together.

Oglopf:

https://osc.github.io/ood-documentation-test/add-globus-example/

**Add your description here**
Fixes #948

### DIFF
--- a/source/reference/files/ondemand-d-ymls.rst
+++ b/source/reference/files/ondemand-d-ymls.rst
@@ -818,7 +818,7 @@ Configuration Properties
 .. _globus_endpoints:
 .. describe:: globus_endpoints (Array<Object>, null)
 
-  Add a Globus button to the file browser that opens the current directory
+  Add a Globus button to the file browser. The button opens the current directory
   in the Globus transfer web app.
 
   Note that ``endpoint_path`` is the path that Globus will initialize to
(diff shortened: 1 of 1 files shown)


source/reference/files/ondemand-d-ymls.rst
```diff
@@ -851,6 +851,16 @@ Configuration Properties
...
 
+    In this example with multiple endpoints, suppose you use the file browser to look at your file system. 
+    If you were to select the ``/home`` path to see the contents and then click the Globus button, you would 
+    see the endpoint associated with ``path: "/home"`` in the Globus interface.
+
+    If you then go back to the OOD file browser and select the ``/project`` path and then click the Globus button, 
+    you would now see the endpoint associated with the ``path: "/project"`` endpoint in the Globus interface.
+
+    The gist here is you *select the path on the filesystem in the file browser app first* then 
+    click the Globus button to show the corresponding endpoint in the Globus interface in a newly opened tab.
```
johrstrom: This section probably just needs to be moved up to line ~821 as it describes the feature even with one endpoint.

I would also suggest using the word `local` a lot to reinforce you're browsing the `local` filesystem in OOD and that the globus endpoints correspond to that `local` filesystem path.

Oglopf: Yeah, I wondered since it does describe both. I'll push it up, plus that will hopefully get the user to read the behavior first rather than as an after thought like I have it.

For the use of `local` I wanted to really try to connect what they are setting in the config with what they see. I'd rather not use `local` as i think that may confuse people. As in we'd get a discourse asking "where to set local?" type thing. I'll mess with the wording a bit though and see what I can do.

johrstrom: > For the use of local I wanted to really try to connect what they are setting in the config with what they see.

What they see is a `local` file system. 

So I'm suggesting something like this:
```
If you were to select the ``/home`` path
```
becomes
```
If you were to navigate to the the local filesystem path ``/home``
```

johrstrom (approved):
Works for me!

PR #982, opened 2024-07-15
removed old issues, added some security pages

matt257:
remove a section with old issues and add a security section


source/issues/overview.rst
johrstrom, line 1: Can we remove this page as a separate PR? I'd love to say we have no known issues, but alas, we kinda do.

source/security.rst
```diff
@@ -0,0 +1,38 @@
...
+This document provides an overview of the security framework implemented in Open OnDemand, detailing security features, general policies, and procedures.
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Security Topics
+
+   security/vulnerability-management
+   authentication/overview 
+   how-tos/monitoring/logging
+   customizations
```
johrstrom, line 18: It's not clear to me why these are here. They seem to fill up the page, but it's not immediately clear how, say customizations, are related to security?

Oglopf: Logging and Auth are paramount in security, leave them.

Customization I can see as we have things about disabling users in there and file uploads being set. It's security, don't skimp on info.

source/security.rst
```diff
@@ -0,0 +1,38 @@
...
+
+Security Features
+-----------------
```
johrstrom, line 21: I think the main point here should be more like `Security Concerns`. Here are some off bullet points I'd add/think folks are expecting.

I think folks are trying to answer the question `if I install this, what security concerns do I need to account for?`.  This page should answer that.

The good:

* PUN (Per user Nginx) architecture. The web servers that the user will ultimately interact with use are processes' ran by that user. That is, the `root` user does not serve requests. Only non-root users serve requests and user request are always routed to the nginx of that same user. This ensures that things like file accesses (or any action taken by the webservers) are always made under the non-root user.
* Apache authentication is required. The type of authentication scheme is up to the individual site. Basic authentication mechanisms like Basic or LDAP are discouraged and undocumented because they are insecure. 

The bad:
* HTTP only traffic to origin servers (compute nodes). When running applications like Jupyter, apache will proxy traffic back to the origin server (Jupyter on a compute node) over plain http traffic.  There's ongoing work to make this https traffic.

source/security.rst
johrstrom: > Logging and Auth are paramount in security, leave them.

Authentication is required either way - I can setup basic auth and call it a day, but that doesn't mean it's secure. And just having it in a ToC doesn't tell me that.

Counterintuitively, the more text you have on a page, the less any of it is read.

matt257: I think at least the below are related to security. Maybe I can update the list and fold the whole "Security Topics" somehow so the links aren't in the way.

Announcements
Message of the Day (MOTD)
Overriding Pages
Add URLs to Help Menu
Configuration Profiles
Changing the Navigation bar
Interactive Apps Menu
Set Upload Limits
Set Download Limits
Block or Allow Directory Access
Disabling Users
Set Default SSH Host
Set SSH Allowlist
Set OOD SSH Port
Shell App SSH Command Wrapper
Fix Unauthorized WebSocket Connection in Shell App
Job Composer Script Size Limit
Hiding Job Arrays
Custom Error Page for Missing Home Directory on Launch
Customize Text in OnDemand
Disk Quota Warnings on Dashboard
Balance Warnings on Dashboard
Maintenance Mode
Disable Host Link in Batch Connect Session Card
Set Illegal Job Name Characters
XDMoD Integration
Accessing Remote File Systems
Cancel Interactive Sessions
Custom Pages
Support Tickets

PR #985, opened 2024-07-16
add basic security pages

matt257:
https://osc.github.io/ood-documentation-test/add-basic-security-pages/

--------------------------------------------------------------------------------

add basic security pages
incorporated feedback from my last attempt


source/security.rst
```diff
@@ -0,0 +1,48 @@
...
+
+Introduction
+------------
+
+This document provides an overview of the security framework implemented in Open OnDemand, focusing on practical security concerns administrators need to consider when installing and managing the platform.
+
+Security Concerns
+-----------------
+
+**The Good:**
```
johrstrom: I know I gave you this phrase in the other comment, but we likely need something better.

matt257: hmm, maybe "State of Security" ? the section could be updated as security concerns are found and develop

source/security.rst
```diff
@@ -0,0 +1,48 @@
...
+-------------------
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Security Topics
+
+   security/vulnerability-management
+   authentication/overview
+   how-tos/monitoring/logging
+   customizations
```
johrstrom, line 50: I'm not in love with blink links. I mean - yes we log, but what's the context here, is that good or bad? Does it need updated, does the default work well? Yes we have customizations, but are they secure/insecure?

I think what we need is more up the page where the other considerations are. Something like

* Out of the box OnDemand will let users explore the file entire system. The web server runs as the users, so it will still adhere to FACLs when attempting to view files. However, you set an allowlist to only show certain file paths or disable this altogether. (with links to the given configurations)

It's stuff like that - describing some behaviors that may be unwanted and specific links on how to control that.

We can keep this ToC here - but I feel like we need to be more direct with the reader and not just hope they explore and find the relevant info (and make the relevant connections) themselves.

source/security.rst
johrstrom: Maybe it's just one section called `considerations`?

johrstrom: Oh, you know I meant those 2 headings **The good** and **The bad** - maybe it's just one section? Or maybe `advantages` and `limitations`? Then maybe another section called `controls` for things like file browsing that you can change after you install?

source/security.rst
matt257: > ... yes we log, but what's the context here, is that good or bad? 

the context is logging works as 'how-tos/monitoring/logging' describes.

I don't think I can tell people if a feature is good or bad. I think I can only remove doubt about what a feature is, so people can be convinced whatever judgements they make (if they make any) are just. 

---

If there are specific situations you think should be documented I can open up issues and work on writing something up for them and getting them added. 

i'll try to update the 'considerations' from your comment

source/security.rst
johrstrom: I think what you have is a good start. Can you fix the conflict so we can see how it's deployed and looks.

matt257:
https://osc.github.io/ood-documentation-test/add-basic-security-pages/security.html

working now

johrstrom (approved):
OK, this is a good start. We can touch it up as we go.

matt257:
thanks! that sounds like a good plan to me.

PR #981, opened 2024-07-15
Add block code for multi shorcut directories

garadar:
Add shorcut for multiple directories

**Modify this link to include the branch name, and possibly the page this PR modifies**:

https://osc.github.io/ood-documentation-test/latest/

## description 
Add block code for multi directories. Hope it will help

PR #986, opened 2024-07-16
fix user contributed example

johrstrom:
https://osc.github.io/ood-documentation-test/fix-user-contrib/

#981 had some formatting issues - this fixes those issues.


johrstrom:
This is being fixed in another PR.

PR #987, opened 2024-07-16
add a controls section here

johrstrom:

https://osc.github.io/ood-documentation-test/security-controls/

This tweaks the security.rst slightly just to provide some extra context around things that folks may want to control (or know that they can control) when installing OOD.

### DIFF
--- a/source/security.rst
+++ b/source/security.rst
@@ -37,8 +37,6 @@ that some centers may want to change or disable altogether.
 Additional Information
 ----------------------
 
-- **Monitoring and Logging**: Comprehensive logging mechanisms are integral for security audits and incident response. Detailed guidelines and settings for these features can be found at :ref:`logging`.
-
 - **Vulnerability Management**: Active management of security weaknesses includes regular updates and patches. Detailed processes and current security advisories are available at :ref:`vulnerability-management`.
 
 - **Security Audits**: The platform undergoes periodic security audits by Trusted CI, the NSF Cybersecurity Center of Excellence. Summaries of these audits are available, with the latest report accessible `here <https://openondemand.org/sites/default/files/documents/Trusted%20CI%20Open%20OnDemand%20Engagement%20Final%20Report%20-%20REDACTED%20FOR%20PUBLIC%20RELEASE%20210712_0.pdf>`_.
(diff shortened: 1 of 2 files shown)


source/security.rst
```diff
@@ -23,8 +23,19 @@ Limitations
...
+These are things the the out of the box OnDemand installation will provide
+that some centers may want to change or disable altogether.
+
+- **File Access**: OnDemand lets users navigate the file system. While file permissions
+  limit what a user can view and navigate to, some centers may want to limit this even further.
+  One option is to :ref:`set-file-allowlist` to limit what directories users may navigate to.
+
+
+Additional Information
+----------------------
```
johrstrom, line 38: I'm not sure about this heading. Happy to change it. Don't really know the purpose of this section, so can't really speak to it.

source/security.rst
```diff
@@ -23,8 +23,19 @@ Limitations
...
------------------
+Controls
+^^^^^^^^
+
+These are things the the out of the box OnDemand installation will provide
+that some centers may want to change or disable altogether.
+
+- **File Access**: OnDemand lets users navigate the file system. While file permissions
+  limit what a user can view and navigate to, some centers may want to limit this even further.
+  One option is to :ref:`set-file-allowlist` to limit what directories users may navigate to.
```
johrstrom: This addition was really what I was getting at in comments in the previous PR.

Instead of just blind links to _all_ customizations - we have specific callouts to specific behaviors that an initial installation may want to change. With a link to that specific configuration.

matt257: Thanks for explaining, that looks good to me.

source/security.rst
matt257: sure, change it to whatever you'd like! I'll remove the monitoring and logging line.

johrstrom: OK - if your'e removing those 2 maybe I can just remove the whole section and move the line about vulnerability reporting somewhere else, maybe nearer the top of the page?

matt257: If you think that will be better please do. I was trying to put info that could be useful for security somewhere, but i'm not sure about the order. i think its a good idea for vulnerability reporting to be mentioned near the top

PR #989, opened 2024-07-17
add info about disabling apps to customizations

johrstrom:

https://osc.github.io/ood-documentation-test/disable-apps/

This adds information about how to disable or limit apps in the customization section. I see now that this is a bit duplicate of what's in app sharing, but I feel like it's more discoverable here. Plus a much more common use case as few sites enable app sharing and never even navigate to that page.


matt257 (approved):
thanks for updating it, it looks good to me!

PR #994, opened 2024-08-03
Update shibboleth example removing deprecated ood_auth_map.regex

ticklejw:
replace with user_map_match config item

related to https://github.com/OSC/ondemand/pull/729


johrstrom (approved):
Thanks!

PR #1062, opened 2025-01-03
Update allowlist.inc

moffatsadeghi:
updating grammar

https://osc.github.io/ood-documentation-test/moffatsadeghi-idkbranch

### DIFF
--- a/source/release-notes/navconfig.inc
+++ b/source/release-notes/navconfig.inc
@@ -1,5 +1,5 @@
 The initializers used to modify the navigation bar and the class
-``NavConfig`` is deprecated in 3.0 and will be removed in 4.0.
+``NavConfig`` are deprecated in version 3.0 and will be removed in version 4.0.
 The 3.x series will continue to support this, but sites should use the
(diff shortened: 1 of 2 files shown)


source/release-notes/navconfig.inc
```diff
@@ -1,7 +1,7 @@
 The initializers used to modify the navigation bar and the class
-``NavConfig`` is deprecated in 3.0 and will be removed in 4.0.
+``NavConfig`` are deprecated in version 3.0 and will be removed in version 4.0.
```
johrstrom, line 2: Are you sure this is plural? The sentence is referencing a single `[the] class Navconfig`.

johrstrom: 🤦‍♂️ it's the plural `intialiazers` that this is tied to. Disregard!

moffatsadeghi: All good! I had to read it a few times because something was off but I couldn't see it at first! Thanks!

PR #1081, opened 2025-02-20
avoid race condition in Lmod spider json file

### LINKED ISSUE #4173: `auto_modules` is sometimes empty
I don't have a way of reproducing this, but twice now on my system we have had the value for `context.auto_modules_foobar` be an empty string in a batch connect job. Restarting web server fixes it. Any ideas on how this could happen?

simonLeary42:
fixes https://github.com/OSC/ondemand/issues/4173


simonLeary42:
https://manpages.debian.org/testing/moreutils/sponge.1.en.html

johrstrom:
Can we use a temp file and `mv` instead? I don't have `sponge` on my EL machine by default so I imagine lots of other folks won't either.

PR #1085, opened 2025-03-04
Update systemd.rst

adamboutcher:
Include an example systemd cluster config and explains some options.

### DIFF
--- a/source/installation/resource-manager/systemd.rst
+++ b/source/installation/resource-manager/systemd.rst
@@ -3,8 +3,55 @@
 Systemd
 =======
 
-Coming in 2.1!
-
 Open OnDemand can use SystemD timers to schedule a 'job'.  Sites can use this adapter for short
 lived jobs or to use resources outside of your compute pool (like login nodes).
+
+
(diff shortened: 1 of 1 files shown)


source/installation/resource-manager/systemd.rst
```diff
@@ -3,8 +3,55 @@
...
+``timeout``
+  is used to set a 'walltime' after which the job is killed.
+``debug``
+  enable debugging for the cluster.
+``strict_host_checking``
+  will force SSH to use strict host checking.
+``ssh_keyfile``
+  If you want to provide a non-standard ssh identify file (equivalent to the -i option in SSH).
+
+.. note:: There are many recipes for managing processes on arbitrary Linux hosts, and we will be exploring others in future development of this adapter.
```
johrstrom: I don't know if this is applicable. I think this adapter will always use `systemd` so future development for this adapter specifically is a bit limited.

adamboutcher: Removed

johrstrom:
Thanks!

PR #1086, opened 2025-03-04
Update form.rst

adamboutcher:
Changes the form docs to highlight that the min/max restrictions on the number field is only client side validated.


johrstrom:
Thanks!

PR #1088, opened 2025-03-11
Update README.md

moffatsadeghi:
https://osc.github.io/ood-documentation-test/moffatsadeghi-patch-1

Remove list of versions besides latest and dev
Listing old versions makes the page seem very outdated

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,13 +18,7 @@ Or select your version:
 
 - [Development](https://osc.github.io/ood-documentation/develop/)
 - [Latest Stable](https://osc.github.io/ood-documentation/master/)
-- [1.6](https://osc.github.io/ood-documentation/release-1.6/)
-- [1.5](https://osc.github.io/ood-documentation/release-1.5/)
-- [1.4](https://osc.github.io/ood-documentation/release-1.4/)
-- [1.3](https://osc.github.io/ood-documentation/release-1.3/)
-- [1.2](https://osc.github.io/ood-documentation/release-1.2/)
-- [1.1](https://osc.github.io/ood-documentation/release-1.1/)
-- [1.0](https://osc.github.io/ood-documentation/release-1.0/)
+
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,13 +18,7 @@ Or select your version:
 
 - [Development](https://osc.github.io/ood-documentation/develop/)
 - [Latest Stable](https://osc.github.io/ood-documentation/master/)
```
johrstrom: This should actually be `latest` and not `master`.

README.md
```diff
@@ -18,13 +18,7 @@ Or select your version:
...
-- [1.6](https://osc.github.io/ood-documentation/release-1.6/)
-- [1.5](https://osc.github.io/ood-documentation/release-1.5/)
-- [1.4](https://osc.github.io/ood-documentation/release-1.4/)
-- [1.3](https://osc.github.io/ood-documentation/release-1.3/)
-- [1.2](https://osc.github.io/ood-documentation/release-1.2/)
-- [1.1](https://osc.github.io/ood-documentation/release-1.1/)
-- [1.0](https://osc.github.io/ood-documentation/release-1.0/)
```
johrstrom, line 27: These are all real - though I don't know if we want them documented. 

You can see older versions of documentation here (at the very bottom left of the navigation panel), though it appears we're missing 3.1 and 4.0, so I can get on that shortly.
image

moffatsadeghi: The intention was to clean this up - I think it is a bit redundant to list every version of documentation we have - rather just point people to dev and latest.  That way we do not have to constantly update this page and have a long list of all the documentation versions to manage. I can edit the PR to reflect this though

johrstrom: Works for me!

moffatsadeghi:
I was providing an edit for you based on your feedback lol.  Will submit another PR

PR #1090, opened 2025-03-12
Update README.md for compromise

moffatsadeghi:
Found a middle ground between feedback from Jeff and how I think this section could improve.  Directions provided to navigate to past releases and direct links to dev and latest.  also updated all mentions of master


johrstrom (approved):
Thanks!

PR #1091, opened 2025-03-17
Fix type key formatting

raj76:
https://osc.github.io/ood-documentation/latest/
Page: source/customizations.rst

Fixes the formatting of the type key.


johrstrom (approved):
Thanks!

PR #1093, opened 2025-03-17
Update installation.rst - Remove FIXME link

genericdata:
https://osc.github.io/ood-documentation-test/latest/installation.html

remove FIXME place holder link. I don't see an actual link, but if link is required I believe it would https://osc.github.io/ood-documentation-test/latest/reference/files/ood-portal-yml.html

#jam

### DIFF
--- a/source/installation.rst
+++ b/source/installation.rst
@@ -16,7 +16,7 @@ means that it will need:
 
 Adding :ref:`Open OnDemand SELinux policies <modify-system-security>` is optional for RHEL systems.
 Open OnDemand, by default, expects Apache to have SSL enabled by :ref:`securing your Apache <add-ssl>`.
-Sites may reconfigure their deployment to allow for plain text traffic see FIXME-LINK-NEEDED.
+Sites may reconfigure their deployment to allow for plain text traffic.
 
 .. toctree::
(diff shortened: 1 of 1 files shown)


source/installation.rst
```diff
@@ -16,7 +16,7 @@ means that it will need:
 
 Adding :ref:`Open OnDemand SELinux policies <modify-system-security>` is optional for RHEL systems.
 Open OnDemand, by default, expects Apache to have SSL enabled by :ref:`securing your Apache <add-ssl>`.
-Sites may reconfigure their deployment to allow for plain text traffic see FIXME-LINK-NEEDED.
+Sites may reconfigure their deployment to allow for plain text traffic.
```
johrstrom: Thanks for the addition - can we clean this up? There's a technical reason why we **needed** SSL for OOD to work properly, the FIXME-LINK-NEEDED was (IIRC) supposed to be a guide to bypassing this.

Now, for a few versions, this point is moot. While sites should run SSL, OOD does not strictly require it as it has in the past.

Which is to say - I think all we need is the first sentence, every thing is isn't actually accurate anymore.

genericdata: Updated.

PR #1097, opened 2025-03-17
Display current version properly at the bottom left corner

rwxayheee:
Hi I'm participating the jam. Thanks for coordinating this event! 

This change applies to most any pages in the documentation. 

### before

Before, `current_version` wasn't specified because it's not in the `html_context` in `source/conf.py`: 

https://github.com/OSC/ood-documentation/blob/f3bd2e0eff88c18ba070b169811f9ecdaf111761/source/_templates/versions.html#L1-L6

So the version was not display as expected: 
Screenshot 2025-03-17 at 3 11 58 PM

### after

`current_version` was added to the `html_context` in `source/conf.py` and points to `release` in the same file. 

A preview from local build: 
Screenshot 2025-03-17 at 3 11 47 PM

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -153,6 +153,7 @@ html_context = {
     'github_repo': 'ood-documentation',
     'github_version': os.environ.get('OOD_BRANCH', 'develop'),
     'conf_py_path': '/source/',
+    'current_version': release,
     'versions' : [
         ['latest', '/latest/'],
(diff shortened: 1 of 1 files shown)


johrstrom (approved):
LOL I've never noticed that that is supposed to be populated. I guess I knew it was a little wonky, but never really looked into it.

Thanks!

PR #1101, opened 2025-03-18
Replace unfriendly keys for navbar_type

genericdata:

The original text says the default is inverse, but you can set it to default if you want other. This is confusing. In the code linked, we have additional more straight forward options of `dark` instead of `inverse` and `light` instead of `default`

https://github.com/OSC/ondemand/blob/4ec6ae009e6886c7d22dfbe16e131d8723cbf979/apps/dashboard/app/models/user_configuration.rb#L93-L107

I also removed URL for bootstrap 3.3, which is outdated. I don't believe a link defining dark and light is required. Also, one less URL to keep up with.


johrstrom (approved):
Yes please and thank you!

PR #1102, opened 2025-03-18
Shorten navbar_type feature name

genericdata:

Compared to the rest of the features in the table I believe we should shorten to the feature rather than define it. Also, it has a period at the end.


johrstrom:
Sure.

PR #1107, opened 2025-03-19
fix env file location for OOD_DOWNLOAD_DIR_MAX

genericdata:

The path is wrong. Fix path to where to set download max size variable.


genericdata:
First noted in

> @johrstrom I see you have updated the docs to use the right path `/etc/ood/config/apps/dashboard/env` but there is still an error in the docs using `/etc/ood/config/apps/files/env `
> 
> 
> 
> ```
> 
> For example, to set the limit to 5 GB, you can add the following line to the /etc/ood/config/apps/files/env file:
> 
> 
> 
> OOD_DOWNLOAD_DIR_MAX=5368709120
> 
> ``` 

 _Originally posted by @pescobar in [#2927](https://github.com/OSC/ondemand/issues/2927#issuecomment-1678710103)_

Issue #977, opened 2024-07-09
Update "Install Other Interactive Apps" section for RStudio


moffatsadeghi:
In the "Install Other Interactive Apps" document, we received feedback in an Office Hours call that the word "hybrid" is unclear in the "RStudio Server" row.  OSC Staff on the call was unsure what "hybrid" RStudio meant as well.  Two indications that our wording for this row should be updated.

OSC now uses BubbleWrap for RStudio Studio and therefore, the documentation should be updated from "hybrid" to "BubbleWrap" as one suggestion.  Another suggestion could be that we expand the definition of hybrid and include the use of "BubbleWrap."  Either way, the documentation should be updated to reflect possible outdated and unclear information.  

Link to the article: https://osc.github.io/ood-documentation/latest/install-ihpc-apps.html

moffatsadeghi:
This issue has been completed in PR #1132

PR #1132, opened 2025-05-20
Update install-ihpc-apps.rst

moffatsadeghi:
Update this file based on #977

### DIFF
--- a/source/install-ihpc-apps.rst
+++ b/source/install-ihpc-apps.rst
@@ -109,7 +109,7 @@ interactive apps that are currently deployed at OSC and other contributing insti
    * - RStudio Server
      - OSC
      - https://github.com/OSC/bc_osc_rstudio_server
-     - BubbleWrap
+     - hybrid (Bubblewrap)
      - rnode
    * - 
      - University of Utah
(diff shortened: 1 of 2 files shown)


source/install-ihpc-apps.rst
```diff
@@ -109,7 +109,7 @@ interactive apps that are currently deployed at OSC and other contributing insti
    * - RStudio Server
      - OSC
      - https://github.com/OSC/bc_osc_rstudio_server
-     - hybrid
```
johrstrom, line 112: Should it be all lowercase? That's the command, `bubblewrap` all lower case.

moffatsadeghi: Oooh, good question.  
https://misile00.github.io/notes/Bubblewrap 
I might go with Bubblewrap actually because it isn't necessarily the command, rather the technology. What do you think about that?

johrstrom: Sure

johrstrom: But it's misspelled, so the CI is failing. 

image

Can you add it to this file?

https://github.com/OSC/ood-documentation/blob/latest/source/spelling_wordlist.txt

moffatsadeghi:
Done :) 
Updated both files based off of feedback

Issue #1035, opened 2024-12-04
Authentication Page Restructure


Oglopf:
While looking at some auth solutions for OOD I noticed some issues with the authentication page's layout and information:
https://osc.github.io/ood-documentation/latest/authentication.html

What is a Protocol vs. a Solution:
  - **Protocols**: protocols like OIDC/OAuth 2.0, CAS (protocol), SAML, and Kerberos define the standards for how authentication and/or authorization are handled. They specify the format and process for exchanging authentication information between systems.
    - Some protocols are defined by RFCs (e.g., OIDC/OAuth 2.0, CAS, and Kerberos), while others are defined by external standards bodies, such as OASIS for SAML.
  - **Solutions**: solutions like Shibboleth, Dex, Keycloak, and CAS (solution) are software implementations or platforms that use these protocols. These solutions act as identity providers (IdPs), service providers (SPs), or intermediaries to enable authentication and authorization *using one or more protocols.*

With these definitions the following issues become apparent:
1. Confusing Hierarchy. The way the documentation presents these topics seems to mix and match protocols and solutions on the same level:
- OIDC is a protocol, but Shibboleth and CAS as _solutions_ are specific systems or software yet exist at the same level as OIDC.
- OIDC has multiple sections for different solutions for Dex and Keycloak, but these should both just be under an OIDC protocol section as solutions.
2. Inconsistent Structure. 
- The way Dex and Keycloak are detailed under OIDC implies they are direct implementations of the OIDC protocol, which makes sense. 
- But then Shibboleth and CAS appear at the same level without this kind of breakdown or clarifying their roles:
  - Shibboleth, while most commonly associated with SAML, **can also act as an IdP** for OIDC or OAuth2 in some setups. This flexibility should be acknowledged, but its primary association with SAML warrants placing it under a SAML section, maybe with a note about its broader capabilities.
  - Similarly, CAS (solution) follows the CAS protocol but is presented in a way that blurs the line between the solution and protocol.

The documentation could be more intuitive and helpful if it followed a more consistent structure, like this:

1. Protocols Overview:
  - __OpenID Connect (OIDC)__
    - General explanation of OIDC
    - Solutions: 
      - __Dex__
      - __Keycloak__
  - __SAML__
    - General explanation of SAML
    - Solutions: 
      - __Shibboleth__
      -  __ADFS__ with `mod_auth_mellon`
  - __CAS__
    - General explanation of CAS protocol
    - Solution: 
      - __CAS__
[...]

PR #1149, opened 2025-07-02
First restructuring pass for auth section

Oglopf:

https://osc.github.io/ood-documentation-test/auth-page-restruct-1035/

**Add your description here**
First pass at restructuring some of the auth page to help with the flow and emphasis on parts that _must_ be done which users have a history of missing (like user mapping).

Part of issue #1035


moffatsadeghi:
ToC is the last piece to tie up - since hitting some funkiness with the rendering.  Discussed and walked through each change.  Looks fantastic on first walkthrough!

Oglopf:
I think I have the ToC finally working correctly and all links seem to be correct. I'm still clicking through some pages to check I haven't missed anything yet.

This makes the landing page for the auth very lean but it is clear what you need to do and i removed a lot of redundancy in wording and warnings and merged some things. The idea is the page is incredibly streamlined so that it is hard to miss what is required, and it's now designed to let you "fall through" the docs section of the ToC then spit you back out to a "Known Auth Solutions" page where the user can select their auth, rather than just being dumped right into OIDC as previously.

I did some serious rework on the overview, to the point i removed the file and even the old authentication page, split the authentication between the overview and the solutions, then folded the overview into that new ood-auth page as it felt like 3 pages all said the same thing in 3 different ways and this destroyed the nesting of the ToC to navigate easily.  So, it's a big change. Hopefully it improves things.

There's still room for improvement, but I am much happier with the logical flow and semantics for now.

Oglopf:
I also reworded the "Insecure Options" page to call out the reason so we don't have to explain it constantly along with making it a bit more severe in the warning of an insecure prod instance.

source/ood-authentication.rst
moffatsadeghi, line 1: I really like that you have titled this OOD Documentation for clearer communication, but the URL needs to stay https://osc.github.io/ood-documentation/latest/authentication.html so that when people are configuring OOD and they get redirected to the documentation, they don't now get a 404 page.  

https://github.com/OSC/ondemand/blob/4acd0a2a3ac3cb46787f757005255d1890be10f2/ood-portal-generator/templates/no-auth.conf.erb#L11

Could you keep the title of the section the same but revert back the URL to the original?

moffatsadeghi: Maybe a more clearer way I could could have put it:

The heading and the text can still have "OOD Authentication" but the URL / maybe the file name needs to keep "Authentication"

moffatsadeghi (changes requested):
This is absolutely awesome, Travis, and I want to get this into our docs soon.  Right now, these changes are in develop, which means they will not get pushed out until 4.1 timeframe because I've confirmed there are several other things waiting in develop that should not go to latest because of how they are tied to future releases.  Is it possible for you to move these changes to latest so we can get it into the docs way sooner? Sorry that I did not mention having you doing this in latest sooner :(

source/ood-authentication.rst
johrstrom: >  but the URL needs to stay https://osc.github.io/ood-documentation/latest/authentication.html 

Yes I believe it's imperative we keep the URL for compatibility with previous versions.

Travis, Emily mentioned to me you might need some help with her suggestion to change the URL back to `authentication.html` from `ood-authentication.html`. If that's the case, I can help walk you through the process and why sphinx is throwing errors. Alternatively I'm happy to provide a diff that you can apply or open another PR against this or any other branch.

PR #1151, opened 2025-07-22
Update Dockerfile URL

genericdata:
https://osc.github.io/ood-documentation-test/latest/installation.html

This [pull request](https://github.com/OSC/ondemand/commit/dd8d917ce65f5c3b9637061bec1d702086ae0e12#diff-2604174ad4aece255b429040694aee0bc73b8b951ffacef5b9968935d59a4ea2) moved Dockerfile to Dockerfile.example. This update fixes the URL and filename on the installation page.

### DIFF
--- a/source/installation.rst
+++ b/source/installation.rst
@@ -32,9 +32,9 @@ Building From Source
 Building from source is left as an exercise to the reader. 
      
 It's not particularly difficult to build the code, but installing it with all the various files is. Should you be interested, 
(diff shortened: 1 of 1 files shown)


johrstrom:
Thanks! Though, now that I think about it, that Dockerfile doesn't build from the source code anymore, so that whole section likely needs to be removed.

That said - I merged it because even if the section is slightly off, it may as well point to a correct URL.

PR #1153, opened 2025-07-24
document usage of `placeholder` attribute

simonLeary42:
as a workaround for https://github.com/OSC/ondemand/issues/3636, I display the default value in the placeholder, and in `submit.yml.erb` I use those default values for fields left blank. I also made sure not to use the `required` attribute, and added a message to `form_header` saying that it's OK to leave fields blank.

### DIFF
--- a/source/how-tos/app-development/interactive/form.rst
+++ b/source/how-tos/app-development/interactive/form.rst
@@ -475,11 +475,11 @@ are:
           placeholder: null
 
      Example
-       Display default value in placeholder
+       Display recommended value in placeholder
 
        .. code-block:: yaml
 
-          placeholder: "Default: 4"
+          placeholder: "Recommended: 4"
 
 .. describe:: pattern (String, null)
 
--- a/source/how-tos/app-development/interactive/form.rst
+++ b/source/how-tos/app-development/interactive/form.rst
@@ -465,7 +465,7 @@ are:
 
 .. describe:: placeholder (String, null)
 
-     help text that appears inside the field when it is empty
+     Example or hint text that is shown inside the field with low contrast when it is empty.
 
      Default
        No help text appears in input field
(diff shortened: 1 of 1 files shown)


source/how-tos/app-development/interactive/form.rst
```diff
@@ -463,6 +463,24 @@ are:
             - `blue`
             - `green`
 
+.. describe:: placeholder (String, null)
+
+     help text that appears inside the field when it is empty
```
johrstrom: Can you make this a proper sentence with capitalization and punctuation. 

Also I worry about `help text` given there's a `help` field. I wonder if there's a better phrase here? I don't know that there is, but I just wonder if we have an alternative here? 

Maybe MDN has something for us? They use the word `hint` but I'm now thinking maybe `example` could work?

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/placeholder

simonLeary42: I was copying the writing style from the `help` attribute: https://github.com/OSC/ood-documentation/blob/2f6d15ab9bd81527419c82b2094d9cb1dc7cef51/source/how-tos/app-development/interactive/form.rst?plain=1#L444

Shall I update that one as well?

johrstrom: Yea I'm sure there are lots of this all over the place. You don't have to update older fields now if you don't like. Or you can open a separate PR to fixup the older attributes. I'll leave that to you.

simonLeary42: OK. How's this?

> Example or hint text that is shown inside the field with low contrast when it is empty.

I'm not sure if this qualifies as a proper sentence since it doesn't have a subject and a predicate.

johrstrom: Works for me.

PR #1154, opened 2025-07-24
document usage of `reportValidity()`, `setCustomValidity()`

simonLeary42:
I've been trying to make the batch connect form more user friendly, and I found that this short code snippet makes a big difference.


simonLeary42:
confirmed code works: 
image

johrstrom:
I don't think this is the right spot for this. Seems like we need a specific page for `form.js`. This tutorial doesn't use any of these fields.

moffatsadeghi:
@simonLeary42 wanted to follow up on this.  Jeff is not suggesting that this be moved to another repo, rather another under the form.js [section](https://osc.github.io/ood-documentation/latest/how-tos/app-development/interactive/dynamic-form-widgets.html#your-own-form-js
 ) so we can use it in other locations besides this matlab tutorial.  Let us know if you have any questions / concerns! Have a nice weekend :)

moffatsadeghi:
@simonLeary42 just wanted to ping you on this again in case you missed it!

PR #1155, opened 2025-08-01
Increase Rakefile compatibility

Bubballoo3:
Removes the need for WSL to build docs, simplifying the process as described in [https://github.com/OSC/ood-documentation-build/issues/24](https://github.com/OSC/ood-documentation-build/issues/24). Tested on Ubuntu and Powershell, should retain old functionality across other UNIX systems and macs (but not tested)

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -54,11 +54,7 @@ def run_cmd
   if podman?
     "podman run --rm -it -v #{__dir__}:/doc #{image}"
   elsif docker?
-    user_section = ''
-    current_dir = '.'
-    if !windows?
-      user_section = "-u '#{user_group}'"
-    end
+    user_section = windows? ? nil : "-u '#{user_group}'"
     "docker run --rm -it -v \"#{__dir__}:/doc\" #{user_section} #{image}"
   else
     raise StandardError, "Cannot find any suitable container runtime to build. Need 'podman' or 'docker' installed."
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -28,20 +34,32 @@ def image
...
   if podman?
     "podman run --rm -it -v #{__dir__}:/doc #{image}"
   elsif docker?
-    "docker run --rm -it -v '#{__dir__}:/doc' -u '#{user_group}' #{image}"
+    user_section = ''
+    current_dir = '.'
+    if !windows?
+      user_section = "-u '#{user_group}'"
+    end
+    "docker run --rm -it -v \"#{__dir__}:/doc\" #{user_section} #{image}"
```
johrstrom: `current_dir` doesn't appear to be used. 

Can you turn `user_selection` into a ternary like

```ruby
user_selection = windows? : nil ? "-u '#{user_group}'"

Also does single quote `'` not work here? `\"` always looks funny to me.
```

Bubballoo3: Thanks for catching that, I removed the current_dir variable and included the ternary in place of the if statement. The double quotes are necessary as the single quotes (when passed through by exec()) are interpreted as a volume name by docker (on windows). We could use no quotes instead, but double quotes will keep the same protection against spaces in the path.

PR #1156, opened 2025-08-06
Fix warning typo

Bubballoo3:
Fixes minor spelling error in warning message


johrstrom:
Can you supply this against the `latest` branch instead of `develop` so that it'll be published now.

`latest` is the current version - what everyone looks at.
`develop` is the next version where we can add things that aren't public yet (for 4.1, the next version for example)

PR #1162, opened 2025-08-11
Set example TurboVNC directory to the default installation path rather than /usr/local

hsmallbone:
https://osc.github.io/ood-documentation-test/turbovnc-docs/how-tos/app-development/interactive/setup/modify-cluster-configuration.html

**Add your description here**
Using a local path without explicitly denoting it as such is confusing to users. I have set it to the default TurboVNC installation directory - alternative may be to call this out specifically.

### DIFF
--- a/source/how-tos/app-development/interactive/setup/enable-reverse-proxy.rst
+++ b/source/how-tos/app-development/interactive/setup/enable-reverse-proxy.rst
@@ -53,7 +53,7 @@ Requirements
             vnc:
               script_wrapper: |
                 module purge
-                export PATH="/usr/local/turbovnc/bin:$PATH"
+                export PATH="/opt/TurboVNC/bin:$PATH"
                 export WEBSOCKIFY_CMD="/usr/local/websockify/run"
                 %s
               set_host: "host=$(hostname -A | awk '{print $1}')"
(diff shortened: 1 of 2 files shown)


johrstrom:
Since this is a shell script content, maybe we can add a comment like so? Do you think that could be helpful?

```bash
module purge

# correct the PATH to point to the turbovnc installation directory,
# which may differ of your system.
export PATH="/opt/TurboVNC/bin:$PATH"
```

hsmallbone:
How about this?

johrstrom:
Works for me.

johrstrom:
Went to merge this and noticed the commits are a bit off. They're not associated with your github user, so you won't really get credit for them, which is important.

moffatsadeghi:
> Went to merge this and noticed the commits are a bit off. They're not associated with your github user, so you won't really get credit for them, which is important.

Does that matter to you @hsmallbone or can we go ahead and merge it?

hsmallbone:
> > Went to merge this and noticed the commits are a bit off. They're not associated with your github user, so you won't really get credit for them, which is important.
> 
> Does that matter to you @hsmallbone or can we go ahead and merge it?

No all good

PR #1164, opened 2025-08-19
add a note about some schedulers having limited support

johrstrom:
https://osc.github.io/ood-documentation-test/limited-support/

add a note about some schedulers having limited support


source/installation/resource-manager/limited-support.inc
moffatsadeghi, line 1: Any chance we can call this "Limited Support Notice" instead of Note? For more clarity?

moffatsadeghi, line 1: Here are two other suggestions:

This resource manager is not actively maintained by the Open OnDemand developers, either because it was community-contributed or we no longer have access to a test system. As a result, bug fixes from the core team may be limited, but community patches are encouraged.

__________________

This resource manager has only limited support from the Open OnDemand developers. In most cases, this is because it was contributed by the community or the development team no longer has access to a live system for testing.

As a result, the developers may not be able to troubleshoot or fix bugs that arise.
That said, community contributions and patches are always welcome to help improve support.

moffatsadeghi (changes requested):
I was not able to pull your changes down on my GitHub CLI *sigh* but reviewing all the files listed and your https://osc.github.io/ood-documentation-test link, I was able to get a good feel for what this would look like implemented. Thanks for your work on this!!

source/installation/resource-manager/limited-support.inc
johrstrom: I am not sure if I can change the title from `Note`, though we can use other directives. Note that `note` is blue while something like `warning` is yellow, so there are a few we can choose from here.

Though it appears that `admonition` (grey) is the only one I can re-title. 

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html

moffatsadeghi: That's fine then.  We can keep Note then.  Thanks!

moffatsadeghi (approved):
lgtm, thanks for making these changes!

PR #1174, opened 2025-09-18
fixes incorrect spelling  in oidc_redirect_uri reference

johrstrom:
https://osc.github.io/ood-documentation-test/fix-develop/

Supersedes #1173 to fix the develop branch spelling errors.

### DIFF
--- a/source/customizations.rst
+++ b/source/customizations.rst
@@ -599,7 +599,7 @@ connection alive is not factored into this.
 of activity (in milliseconds). After this duration, the connection will be closed
 regardless of activity. Its default is 3600000 milliseconds (1 hour).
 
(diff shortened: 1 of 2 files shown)


moffatsadeghi (commented):
Based off some research, it does appear that URI should be capitalized, but that it should have the hyphen.  
So "Sub-URI" 
One of the data points that I used: https://discourse.openondemand.org/t/initial-install-does-not-display-dashboard/1991/2 

That’s the way Apache and 'mod_auth_openidc' refer to it in their docs. Writing “Sub URI” without the hyphen looks off, because it makes “Sub” and “URI” seem like separate concepts instead of a single directive name.

johrstrom:
Yea it appears that `Sub-URI` is correct spelling too.

johrstrom:
Changed to `Sub-URI` so I'm going to merge.

PR #1185, opened 2025-10-10
Improvements to the support ticket documentation for RequestTracker/ServiceNow

abujeda:
## Description

Added more information on how to use and configure RequestTracker and ServiceNow for the support ticket feature.

Page Modified:
https://osc.github.io/ood-documentation/latest/customizations.html#support-tickets

### DIFF
--- a/source/customizations/support-ticket.inc
+++ b/source/customizations/support-ticket.inc
@@ -36,7 +36,7 @@ For production use, you'll likely want to customize the form, add SMTP settings,
 
 .. raw:: html
 
-   <div style="margin-top: 1.5rem;"></div>
+   <p></p>
 
 The support ticket page has a simple form to gather information that will be submitted to the support system.
 The form can be pre-populated with contextual information in two ways:
--- a/source/customizations/support-ticket.inc
+++ b/source/customizations/support-ticket.inc
@@ -3,54 +3,125 @@
 Support Tickets
 ---------------
 
-The Dashboard now supports sending a support ticket to your institution Help Desk system.
-The feature uses email as the delivery mechanism, but it could be extended to support others.
+The Dashboard enables users to easily submit support requests directly to your institution's Help Desk system.
+This feature provides a customizable form that collects user input and sends the information to the configured system.
 
 To enable this feature, define the settings needed using the configuration property ``support_ticket``.
 Once this configuration object is defined, the ``Submit Support Ticket`` link will be shown in the ``Help`` navigation menu
 and in all interactive session cards.
 
+Quick Start
+...........
+
+The simplest configuration to get started with email-based support tickets:
+
+.. code-block:: yaml
+
+   # /etc/ood/config/ondemand.d/support_ticket.yml
+   support_ticket:
+     email:
+       to: "support@example.edu"
+
+This minimal configuration will:
+
+* Enable the support ticket feature
+* Display the default form with all standard fields
+* Send tickets via email to the specified address
+* Use the default SMTP configuration from Rails
+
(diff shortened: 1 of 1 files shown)


moffatsadeghi:
@abujeda, thank you SO much for this contribution to our docs.  I would usually look at this within a few days, but I have been and will be in and out of the office until early next week.  I have this on my list to review next week.  
Gracias por tu paciencia :)

source/customizations/support-ticket.inc
```diff
@@ -3,54 +3,125 @@
 Support Tickets
 ---------------
 
-The Dashboard now supports sending a support ticket to your institution Help Desk system.
-The feature uses email as the delivery mechanism, but it could be extended to support others.
+The Dashboard enables users to easily submit support requests directly to your institution's Help Desk system.
+This feature provides a customizable form that collects user input and sends the information to the configured system.
```
moffatsadeghi: Maybe "... sends the information to the designated support system."

moffatsadeghi: I think that is a bit more clear than "configured system"

source/customizations/support-ticket.inc
```diff
@@ -3,54 +3,125 @@
...
+
 .. figure:: /images/support_ticket_menu.png
    :align: center
 
-The support ticket page has a simple form to gather information that will be submitted to the support system via email.
-If triggered from a session card, the system will automatically add information about the selected
-session to the body of the email to help troubleshoot the issue.
+.. raw:: html
+
+   <div style="margin-top: 1.5rem;"></div>
```
johrstrom: We shouldn't be telling folks to use inline styles. Instead they should use a class like `mt-1` or similar.

abujeda: I have updated to use an empty `p `tag that is styled with the required margin bottom with the Sphinx CSS.
mt-1 is from Bootstrap and not available in Sphinx. I could not find something similar in the Sphinx  CSS.

This is to make this image and paragraph look good. Otherwise the spacing is too small, like so:
Screenshot 2025-10-24 at 20 38 16

abujeda: This is with the current fix:
Screenshot 2025-10-24 at 20 38 58

johrstrom: 🤦‍♂️ I read that as being something like this below, not an actual HTML element on the page, but rather documentation. That's fine to use inline sytles here, my comment was geared more towards communicating how to build HTML elements to the reader, not building HTML in the actual page.

```html
<div></div>
```

moffatsadeghi (changes requested):
Two small changes and everything else looks great!

source/customizations/support-ticket.inc
abujeda: fixed

source/customizations/support-ticket.inc
moffatsadeghi: @johrstrom so what he originally had was fine after all? Asking because the PR is in great shape for me to merge.

moffatsadeghi:
Only other comment - this is on the develop branch which means it will not show up in our documentation until 4.1 is released. If the PR for something already existing or something for the next release?

abujeda:
The new documentation refers to support ticket integrations with Request Tracker and ServiceNow.
Request Tracker integration is already released in OODv3.x.
Service now will be released in OODv4.1

moffatsadeghi:
> The new documentation refers to support ticket integrations with Request Tracker and ServiceNow. Request Tracker integration is already released in OODv3.x. Service now will be released in OODv4.1

Okay sounds good, thanks for the response and the contribution!!

moffatsadeghi (approved):
Approved!

PR #1190, opened 2025-10-27
update websockify min version to 0.13.0

johrstrom:
https://osc.github.io/ood-documentation-test/websockify-version/

Specify the minimum version for websockify to be `0.13.0` as lower versions have issues.


moffatsadeghi (approved):
A bit outside my technical knowledge, but since we are bumping versions and need a reviewer, I approve.

PR #1197, opened 2025-11-10
Kickoff Project Manager Docs

### LINKED ISSUE #1242: Project Manager section
This needs a good overview in the release notes as well as a dedicated user/setup guide. This should include each of https://github.com/OSC/ondemand/pull/4258, https://github.com/OSC/ondemand/pull/4505, https://github.com/OSC/ondemand/pull/3981.

Bubballoo3:
Fixes #1242. This PR is mainly about nailing down the directory configuration we want for the project manager, specifically the numbered items at the bottom of the page, in addition to explaining the `OOD_SHARED_PROJECT_ROOT` environment variable. The rest is the bare minimum to provide the structure leading up to these requirements, and I have absolutely skipped some key in-between content. I also placed a direct navbar item for `Project Management` for our convenience, mostly because I don't know exactly how we will end up fitting this section in to the documentation structure. While we can discuss the best approaches here, the main focus of this PR is to discuss the PM details so that we have a source of truth we can refer to as we continue development, even if we don't merge this until we document the rest of 4.1.

View the new page at https://osc.github.io/ood-documentation-test/initialize-PM-docs/how-tos/project-management/project-manager.html


source/how-tos/project-management/project-manager.rst
```diff
@@ -0,0 +1,65 @@
...
+to collaborate together they will need a properly configured shared directory,
+meeting :ref:`these minimum permissions <project_manager_shared_project_root>` 
+
+
+.. _project_manager_collaboration:
+
+Collaboration
+-------------
+Collaborative projects need properly configured directories to exist in, which may 
+vary on the types of collaboration you would like to enable. Like other actions in
```
Bubballoo3, line 55: I realize that the 'types of collaboration' referenced here is a little nebulous, but here is what I had in mind. 

Basically, we should support both 'read-only collaboration' (collaborators can view, copy, and execute files, allowing them to run any launchers and workflows) and 'full collaboration' (collaborators have identical permissions to project owner).

I envision certain centers being uncomfortable with 'full collaboration', and so they could configure their directories to never allow non-owning collaborators to edit files. I also envision some project owners wanting to restrict their collaborators from editing specific projects (or limit 'full collaboration' to individuals within their group), so this would be an option in the project settings. 

For these reasons, I would say that we recommend centers create shared directories with full group permissions, giving individual project owners the ability to restrict projects to read-only if they wish. These permission restrictions would only affect the project root (not shared root or group folders), and so would not need additional involvement from an admin.

source/how-tos/project-management/project-manager.rst
```diff
@@ -0,0 +1,65 @@
...
+
+.. _project_manager_collaboration:
+
+Collaboration
+-------------
+Collaborative projects need properly configured directories to exist in, which may 
+vary on the types of collaboration you would like to enable. Like other actions in 
+Open OnDemand, it will operate as the logged in user and never exceed the UNIX 
+permissions of the directories and files it operates on. This means there are several
+different approaches you may want to take depending on your file system's account and
```
Bubballoo3, line 58: We should also put some thought into other approaches to file permissions and how those will cooperate with the PM. (I know some places use our Allowlist functionality, but is this an alternative to group-level management?) Basically we should be minimal with our required permissions in order to remain compatible with whatever strategies people are using, but advocate and pull examples from an OSC inspired format. (Even OSC doesn't currently have these permissions yet, but we should make sure they are comfortable implementing our recommended approach).

source/how-tos/project-management/project-manager.rst
```diff
@@ -0,0 +1,65 @@
...
+easily accessible to others in the ``developers`` group (likewise for ``staff`` and ``students``).
+
+While the Project Manager automates the permissions settings on project-specific folders, these can
+never exceed the permissions of the ``SHARED_PROJECT_ROOT`` or any group-level subdirectory, so it 
+is important to ensure that your directory structure meets these minimum requirements.
+
+#. Any directory in ``OOD_SHARED_PROJECT_ROOT`` and above must have at minimum ``r-x`` permissions
+   for all potential collaborators.
+
+#. Any group directory directly below ``OOD_SHARED_PROJECT_ROOT`` should have ``rws`` permissions
```
Bubballoo3: We can also definitely add a section below explaining what these permissions mean and how to enact them, like `setgid` to get the `s` permission. But this can wait until we nail down all the permissions we need.

source/how-tos/project-management/project-manager.rst
```diff
@@ -0,0 +1,65 @@
...
+
+While the Project Manager automates the permissions settings on project-specific folders, these can
+never exceed the permissions of the ``SHARED_PROJECT_ROOT`` or any group-level subdirectory, so it 
+is important to ensure that your directory structure meets these minimum requirements.
+
+#. Any directory in ``OOD_SHARED_PROJECT_ROOT`` and above must have at minimum ``r-x`` permissions
+   for all potential collaborators.
+
+#. Any group directory directly below ``OOD_SHARED_PROJECT_ROOT`` should have ``rwx`` permissions 
+   for the group
```
Bubballoo3, line 92: In light of discussion, we will try to handle setbit and sticky bit automatically in the project root, as described in https://github.com/OSC/ondemand/issues/4780 and https://github.com/OSC/ondemand/issues/4781

moffatsadeghi:
Putting this comment here about Project Manager documentation in case it gets buried:

"I think some documentation would be great. As someone who largely works in the terminal, it was not immediately clear to me how what I am doing maps to the launcher script, or how I should modify my launcher script to use variables I create. I'll keep flailing away at it though. "

Bubballoo3:
Turns out I was mistaken about the name of the environment variable here, which should have been `OOD_SHARED_PROJECT_PATH`. The content has been updated with this change

source/how-tos/project-management/project-manager.rst
Bubballoo3: In subsequent development we decided not to support 'read-only shared projects' instead shared projects will have 'full collaboration' by default and any modifications from that will have to be effected from the command line.

PR #1198, opened 2025-11-24
Add Bootstrap info

saracook:
https://osc.github.io/ood-documentation/latest/customizations.html#overriding-pages

I had to dig to find the Bootstrap version because it's not documented, so this is an attempt to fix that for future readers.

It's worth explicitly telling people what Bootstrap is and that it is built into OOD, if only to improve the search results.

Currently, a search for 'Bootstrap' in the docs produces outdated release notes that reference previous versions of Bootstrap, and an unexplained reference to Bootstrap alerts in https://osc.github.io/ood-documentation/latest/customizations.html#announcements. Users should be able to get better information for that search.

I'm not claiming that my text is perfect; please wordsmith as needed.

Thanks!
Sara

### DIFF
--- a/source/customizations/overriding-partials.inc
+++ b/source/customizations/overriding-partials.inc
@@ -34,10 +34,10 @@ distributed by the Open OnDemand package.
 
 .. code-block:: html
 
-  <div class="text-center p-4 my-3" >
+  <div class="text-center p-4 my-3 fs-5" >
     My center's custom footer.
   <div>
 
-The [utility classes](https://getbootstrap.com/docs/5.0/utilities/spacing/) for text alignment, padding, and margin in the sample are from Bootstrap, the front-end framework used in Open OnDemand. Using Bootstrap classes in custom overrides, widgets, and pages streamlines development and maintains a consistent user experience. The current installed version is Bootstrap 5.3.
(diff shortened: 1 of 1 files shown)


source/customizations/overriding-partials.inc
```diff
@@ -34,8 +34,10 @@ distributed by the Open OnDemand package.
 
 .. code-block:: html
 
-  <div class="text-center h-5 my-3" >
+  <div class="text-center p-4 my-3" >
```
johrstrom: Can we keep `h5` (and also correct it)?

johrstrom:
Can you rebase this to `latest` instead of `develop`. `latest` is the current content people see, and develop is for the _next_ version whatever that may be. In any case, I'd say this applies to the current release, so I'd like to put it into latest.

source/customizations/overriding-partials.inc
saracook: Oh, that was supposed to be h5! That makes so much more sense than trying to set height to 5, which would be the closest Bootstrap meaning. I was so confused.

Text alignment, padding, and margins are better examples than the header utility class. The header utility classes are a more niche thing that most people won't need to use in their apps.

johrstrom: > Text alignment, padding, and margins are better examples than the header utility class. The header utility classes are a more niche thing that most people won't need to use in their apps.

OnDemand (or maybe just me) uses it a lot to change font sizes. Though looking through the docs there's an actual class `fs-N` for the same. In any case, that was what I was trying to illustrate - how to change the font size. 

Can we add `fs-5` then?

saracook: Perfect! Change is made.

johrstrom (approved):
Thanks for the update!

PR #1253, opened 2025-12-19
Add data help documentation

### LINKED ISSUE #1214: Data-help directive from #4611
Batch connect applications can now use data-help directives to change help messages in https://github.com/OSC/ondemand/pull/4611.

Need to update dynamic form item docs accordingly.

Bubballoo3:
https://osc.github.io/ood-documentation-test/add-data-help-1214/

Fixes #1214 both in the release notes and in the standard docs. Also adds a line to data-label mentioning it's unset behavior. 

I would appreciate any opinions on these changes, as it is a bit of a test run for the rest of the sections I will complete. The main point I am still unsure about is that I used my existing test form, which has unrealistic but semantically rich attributes and options. Should we prefer using realistic values or does this seem more helpful?

### DIFF
--- a/source/how-tos/app-development/interactive/dynamic-form-widgets.rst
+++ b/source/how-tos/app-development/interactive/dynamic-form-widgets.rst
@@ -193,34 +193,27 @@ form element based on the selected option in a select widget.
 
 .. code-block:: yaml
 
-  attributes:
-    group:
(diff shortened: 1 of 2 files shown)


moffatsadeghi:
LGTM!

PR #1308, opened 2026-02-10
Add images to release notes and project manager tutorial

Bubballoo3:

https://osc.github.io/ood-documentation-test/add-release-note-images/

**Add your description here**


moffatsadeghi:
Photos look great! Do we have one we can show for composite job workflows for the PM?

Bubballoo3:
I have one in the project manager tutorial, but I thought the dashboard would be a better candidate for the release note, as it kind of encompasses everything else. Happy to change that or add additional images if you think it would be helpful though

Bubballoo3:
Also do we like the images being placed in the middle of the notes? I think that makes it super clear which image goes with what, but I could be persuaded to move it under the header (and before the note) if we want to avoid splitting the text

PR #1313, opened 2026-02-13
fix nodejs version in install

johrstrom:
https://osc.github.io/ood-documentation-test/fix-41-install/

fix nodejs version in install to be 22 instead of 20.


johrstrom:
Superseded by #1314.

PR #1322, opened 2026-03-03
Correct RHEL10 Installation Dependencies

genericdata:
https://osc.github.io/ood-documentation/latest/installation/install-software.html#enable-dependencies

RHEL 10 install says to install codeready builder for 9 rather than 10.


johrstrom:
Thanks!

PR #1329, opened 2026-04-01
Update project and workflow manager tutorial and documentation

harshit-soora:

- In Project Manager documentation, added information around error reporting feature of workflows.
- Under tutorial, I have added a video showing: Creating launcher, creating workflow, submitting and job status

I have tested the changes and video integration locally. I have also done local spellcheck.


Bubballoo3:
I think if you use the .gif format you should be able to insert it as a plain image and let the HTML figure it out. Did you try that initially and/or could you see if that works before we add new dependencies?

harshit-soora:
I have kept the mp4 version of video into the documentation for safe keeping as gif quality is lower than mp4's

source/conf.py
```diff
@@ -200,7 +200,7 @@
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ['_static']
+html_static_path = ['_static', 'videos']
```
Bubballoo3: It looks like this is not necessary to show the gif. I can handle removing it though

Bubballoo3 (approved):
This looks great! Thanks so much for recording the walkthrough, I think it will help a lot with using the UI

PR #1333, opened 2026-04-10
Adds accessibility fixes as report through pa11y

Oglopf:
**Modify this link to include the branch name, and possibly the page this PR modifies**:

https://osc.github.io/ood-documentation-test/accessibility_fixes/

**Add your description here**
<html><head></head><body><h1>OOD Documentation Accessibility Fixes (WCAG 2.1 AA)</h1>
<h2>Overview</h2>
<p>This document summarizes the accessibility audit and remediation work performed on the Open OnDemand documentation site (<code>osc.github.io/ood-documentation</code>). The goal was WCAG 2.1 Level AA compliance, driven by the April 24, 2026 ADA Title II deadline.</p>
<p><strong>Tooling:</strong> <a href="https://pa11y.org/">pa11y</a> with <code>--timeout 30000 --wait 2000</code> flags (the wait is required to allow stylesheets to fully apply before auditing).</p>
<p><strong>Scope:</strong> All pages in the built Sphinx HTML output.</p>
<p><strong>Starting state:</strong> 185 errors on the index page alone, thousands across the full site.</p>
<p><strong>End state:</strong> Zero errors across the entire site.</p>
<hr>
<h2>Changes</h2>
<h3>1. CSS Overrides (<code>source/_static/css/custom.css</code>)</h3>
<p>All contrast and visual fixes were made via CSS overrides rather than modifying the upstream RTD theme. This approach keeps fixes isolated, survives theme updates, and requires no changes to RST source for visual issues.</p>
<h4>Link contrast (WCAG 1.4.3)</h4>
<p>The default RTD theme link color (<code>#2980B9</code>) fails 4.5:1 contrast against the white page background. All link types were targeted:</p>
<ul>
<li><strong>Body links</strong> (<code>a.reference.external</code>, <code>a.reference.internal</code>) → <code>#237ab3</code></li>
<li><strong>Header anchor links</strong> (<code>a.headerlink</code>) → <code>#237ab3</code></li>
<li><strong>Breadcrumb links</strong> → <code>#237ab3</code></li>
<li><strong>Table cell links</strong> → <code>#1a6fa8</code> (slightly darker to account for table zebra-stripe backgrounds)</li>
<li><strong>Blockquote links</strong> → <code>#1a6fa8</code></li>
<li><strong>Links with bold text wrapping</strong> (<code>a strong</code>) → <code>#1a6fa8</code></li>
[...]

PR #1351, opened 2026-05-01
Adds accessibility fixes per WCAG 2.1 level AA, audited with pa11y

Oglopf:

https://osc.github.io/ood-documentation-test/accessibility_fixes_v2/

**Add your description here**
**Replaces #1333**, rebuilt against current `latest` after the rename of `source/customization.rst` to `source/customizations.rst`. Same scope of changes and same end state (zero pa11y errors across the site), but now cleaned up for an easier merge.

### DIFF
--- a/source/_static/css/custom.css
+++ b/source/_static/css/custom.css
@@ -58,7 +58,12 @@ footer p {
   color: #ffffff !important;
   text-decoration: underline;
 }
+
 .admonition a.reference {
+  color: #2071a5 !important;
+}
+
+.admonition.note a.reference {
   color: #000305 !important;
   text-decoration: underline;
 }
--- a/source/_static/css/custom.css
+++ b/source/_static/css/custom.css
@@ -58,13 +58,6 @@ footer p {
   color: #747474;
 }
 
-/* Admonition title - both class variants */
-.admonition .admonition-title,
-.admonition > .admonition-title {
-  background-color: #367caa;
-  color: #ffffff;
-}
-
 .admonition .admonition-title a {
   color: #ffffff !important;
   text-decoration: underline;
@@ -78,9 +71,7 @@ footer p {
 .wy-nav-side .wy-menu-vertical a {
   color: #ffffff;
 }
-.wy-nav-side .wy-menu-vertical p.caption span.caption-text {
-  color: #ffffff;
-}
+
 .wy-nav-side .wy-menu-vertical li a code,
 .wy-nav-side .wy-menu-vertical li a code span,
 .wy-nav-side .wy-menu-vertical li a code span.pre {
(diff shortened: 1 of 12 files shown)


moffatsadeghi (approved):
Looks good to me

source/_templates/searchbox.html
```diff
@@ -0,0 +1,20 @@
+<div id="searchbox" role="search">
+  <h3 id="searchlabel">{{ _('Quick search') }}</h3>
```
Bubballoo3: Is the 'Quick search' label necessary? It seems to take valuable vertical space, and the searchbar itself is separately labelled below. In the same vein, I think a role='search' region for a single form element is probably extraneous.

source/_templates/searchbox.html
```diff
@@ -0,0 +1,20 @@
...
+  <div class="searchformwrapper">
+    <form class="search" action="{{ pathto('search') }}" method="get" id="rtd-search-form" class="wy-form">
+      <label for="search-input" class="sr-only">{{ _('Search docs') }}</label>
```
Bubballoo3: This seems more complicated than a simple aria-label on the input, is there a reason to place it in its own element?

source/_static/css/custom.css
```diff
@@ -43,4 +43,187 @@ span.c1 {
...
+  color: #000305 !important;
+  text-decoration: underline;
+}
+
+/* Sidebar nav */
+.wy-nav-side .wy-menu-vertical a {
+  color: #ffffff;
+}
+.wy-nav-side .wy-menu-vertical p.caption span.caption-text {
+  color: #ffffff;
```
Bubballoo3: This line makes the headers and items the same color, despite the headers not presenting any contrast issues 

Image

I think the visual distinction between headers and items is pretty important here, so is there another reason to change them that I am missing?

source/_static/css/custom.css
```diff
@@ -43,4 +43,187 @@ span.c1 {
...
+  color: #000305;
+}
+
+.admonition .admonition-title {
+  /* keep your existing title styling */
+}
+
+/* Force readable contrast on admonition body content */
+.admonition {
+  background-color: #ffffff !important;
```
Bubballoo3: I think this is a little harsh (moving straight to white instead of a lighter background) when normal text seems to be fine wrt contrast. A less invasive fix IMO would be to darken the links within notices/warnings, since those are the only elements that currently introduce issues.

Image

Bubballoo3 (changes requested):
Overall I think this looks alright, but there are a few places where I think we are going a little too far with the CSS styles. I haven't dug into each and every new rule, but I have pointed out a few where it seems like we can take a more minimal approach to achieve contrast compliance. 

The same goes for the searchbar, it seems like adding the 'Go' submit button definitely improves the experience for screenreader users, but there are a lot of other modifications there that have debatable impacts (like an additional region or the duplication of labels).

Finally, it looks like the div.version above the searchbar is still below contrast requirements.

Bubballoo3:
I made a few changes addressing the most important comments. It passed pa11y locally, but this should probably be double checked. There are still a few things i'd like to look into tomorrow, but if we run out of time I'd be happy to merge as is

Bubballoo3 (approved):
I have made my changes and am happy with it. It has passed a full pa11y scan on my side, but I will leave open for a little bit so that @Oglopf can corroborate

PR #1353, opened 2026-05-04
add bc_desktop SAFE_PATH to 4.1 release notes

### LINKED ISSUE #1217: Desktop safe path updates from #4703
bc_desktop now removes HOME paths from the PATH in https://github.com/OSC/ondemand/pull/4703.

Need to describe how this changes setup/installation.

johrstrom:
https://osc.github.io/ood-documentation-test/safe-path/

Fixes #1217.  Unfortunately we don't really have a spot for desktops, beyond how to enable them.  Since users can't modify this app (or this fix to the `PATH`), I can only figure this goes in the release notes.

### DIFF
--- a/source/release-notes/v4.1-release-notes.rst
+++ b/source/release-notes/v4.1-release-notes.rst
@@ -361,8 +361,8 @@ Desktop safe clean PATH
 .......................
 
 Previous versions of Open OnDemand have had issues with the ``bc_desktop`` application where users who
-specify ``PATH`` in their ``.bashrc`` (generally through conda environments) launching desktops.
-Indeed, previous versions of Open OnDemand have even attempted to fix this. Now, in 4.1, we beleive this
+specify ``PATH`` in their ``.bashrc`` (generally through ``conda`` environments) launching desktops.
+Indeed, previous versions of Open OnDemand have even attempted to fix this. Now, in 4.1, we believe this
 specific issue is resolved because the desktops will strip anything from the users ``HOME`` from the
 ``PATH`` environment variable when starting the desktops.
 
--- a/source/release-notes/v4.1-release-notes.rst
+++ b/source/release-notes/v4.1-release-notes.rst
@@ -360,10 +360,11 @@ their jobs to know when they start running.
 Desktop safe clean PATH
 .......................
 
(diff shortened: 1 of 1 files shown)


source/release-notes/v4.1-release-notes.rst
```diff
@@ -357,6 +357,15 @@ their jobs to know when they start running.
 .. figure:: /images/interactive-app-notifications.png
    :alt: The 'My Interactive Sessions' page of Open OnDemand, highlighting the switch to enable notifications.
 
+Desktop safe clean PATH
+.......................
+
+Previous versions of Open OnDemand have had issues with the ``bc_desktop`` application where users who
+specify ``PATH`` in their ``.bashrc`` (generally through ``conda`` environments) launching desktops.
```
Bubballoo3: I think this sentence is a bit malformed. Instead of 'where users ... launching desktops' it should be 'where users ... were unable to launch desktops' (assuming that was the impact)

source/release-notes/v4.1-release-notes.rst
```diff
@@ -357,6 +357,15 @@ their jobs to know when they start running.
...
+Indeed, previous versions of Open OnDemand have even attempted to fix this. Now, in 4.1, we believe this
```
Bubballoo3: I think it could be helpful to go into more detail about the previous vs current approaches. I'm not too sure on the details, but looking at the PR it seems something like 'previous versions used a fixed PATH variable that was unaffected by user definitions. In 4.1, we allow custom PATH locations provided they lie outside the user's home directory'. Is that accurate?

source/release-notes/v4.1-release-notes.rst
johrstrom: I see that - how about now?

source/release-notes/v4.1-release-notes.rst
johrstrom: Updated to 

`Indeed, previous versions of Open OnDemand have even attempted to fix this by hard coding a basic ``PATH``,
though this had issues of it's own if binaries were outside of a basic ``PATH``.`

Bubballoo3 (approved):
LGTM!

PR #1354, opened 2026-05-05
add a task for cleaning the build workspace

johrstrom:
In building recently I found that I wanted a way to clean the workspace to make sure what I was looking at was pristine. So this adds a `rake clean` task to clean the workspace inbetween build edits/git checkouts and so on.

Not a document update so there's no link.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -10,6 +10,11 @@ task :build do
   exec cmd
 end
 
+desc "Clean workspace"
+task :clean do
+  exec "#{run_cmd} make clean"
+end
+
 desc "Spellcheck"
 task :spellcheck do
   exec "#{run_cmd} make spellcheck"


Bubballoo3:
This looks fine but I guess I am a bit confused on why you would need this. You would have to rebuild the html between edits/checkouts anyway so why would you run `rake clean; rake build; rake open;` if it is equivalent to `rake build; rake open`?

johrstrom:
> if it is equivalent to rake build; rake open?

I guess I'm not 100% sure that it is. Does `rake build` for sure get rid of what's previously been built?

Bubballoo3:
> Does rake build for sure get rid of what's previously been built?

I believe it scans for your changes and only updates what is necessary (similar to makefiles maybe?). So I suppose you could use this to force a full rebuild instead of 'build out-of-date files (only)' 

For reference with a single file changed `rake build` takes about 5 seconds, while `rake clean; rake build` consistently takes about a minute since it has to recompile a bunch of files that haven't actually changed. So maybe could be useful for some special cases, but overall much less efficient to be doing each time.

It also seems like deleting the build files is already pretty easy to do if you had to, so having it be a command seems a little strange if you aren't anticipating doing it very often. Just some observations, if you have a use case/would just like to have it in easy reach then I don't have a problem

johrstrom:
> It also seems like deleting the build files is already pretty easy to do if you had to, so having it be a command seems a little strange if you aren't anticipating doing it very often. Just some observations, if you have a use case/would just like to have it in easy reach then I don't have a problem

OK cool. Yea I didn't put it as a task dependency or anything, it's just a convenience thing.

Bubballoo3:
Actually since then I have found an application. The build task only looks for changes to the content, but css and (I presume) other non-html changes require a clean to be updated

PR #1358, opened 2026-05-06
Add job composer note

Bubballoo3:
Notes that the job composer was not updated to the same accessibility standard as the dashboard and that compliance-oriented sites should disable it.

### DIFF
--- a/source/release-notes/v4.2-release-notes.rst
+++ b/source/release-notes/v4.2-release-notes.rst
@@ -22,6 +22,10 @@ but we recommend specifically checking the following configurations that are dir
    - The default value has been changed from false to true in v4.2. Ensure that your custom configuration does not override this default.
    - The ``show_all_apps_link`` configuration can be set through the environment variable ``SHOW_ALL_APPS_LINK`` or in your :ref:`ondemand-d-ymls`.
 
+- Disable Job Composer
+   - While basic accessibility improvements have been made to the Job Composer, it constitutes a separate set of webpages from the dashboard, and so was not evaluated as part of the VPAT.
+   - In order to meet AA compliance, we recommend disabling the Job Composer and directing users to the Project Manager, which has been fully updated and evaluated for accessibility.
+   - The Job Composer can be easily disabled for all users by an administrator with the command ``sudo chmod 700 /var/www/ood/apps/sys/myjobs``.
 
 Operating System Support
 ------------------------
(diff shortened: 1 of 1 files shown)


source/release-notes/v4.2-release-notes.rst
```diff
@@ -22,6 +22,10 @@ but we recommend specifically checking the following configurations that are dir
    - The default value has been changed from false to true in v4.2. Ensure that your custom configuration does not override this default.
    - The ``show_all_apps_link`` configuration can be set through the environment variable ``SHOW_ALL_APPS_LINK`` or in your :ref:`ondemand-d-ymls`.
 
+- Disable Job Composer
+   - While basic accessibility improvements have been made to the Job Composer, it constitutes a separate set of pages from the dashboard, files, and other core apps, and so was not evaluated as part of the VPAT.
```
johrstrom, line 26: `was not evaluated as part of the VPAT.` ? Is that right - I'm looking at it

Bubballoo3: I was trying to capture the discussions we had earlier around 'separate set of pages', like why we did not decide to disable it by default. Are you saying that it is fully compliant and the note is unnecessary?

johrstrom: That discussion centered around this criteria that's basically 'same navigation bar for all pages within a set of web pages' - I argued that `/pun/sys/myjobs` is a different 'set of webpages' from `/pun/sys/dashboard`, so we do support that criteria.

https://www.w3.org/TR/WCAG20/#consistent-behavior-consistent-locations

> Are you saying that it is fully compliant and the note is unnecessary?

I don't know if it's fully compliant, but I'm a little worried about saying that it's not in the VPAT at all.

Bubballoo3: > I'm a little worried about saying that it's not in the VPAT at all.

I think that's a valid concern, but it also seems like we should be clear on the scope of the VPAT. In my mind, the VPAT covers the dashboard and core apps contained within it (ie, in that set of web pages), while everything else is out of scope. So while we did put work into improving the Job Composer's accessibility, this isn't recorded or attested anywhere unless we fill out a second VPAT for the set of pages comprising the Job Composer. As a deprecated tool, I would rather make no claims about it and have people disable it, rather than state that it is compliant and then have to fix it if someone reports a gap. 

But I think the deprecation piece is what really saves us here. The Job Composer is deprecated, and since we are not offering long term support for the tool it does not make sense to perform a separate accessibility assessment on it. The only reason we didn't remove it for 4.2 was for the people who are not bound by the WCAG rules, who can then decide for themselves whether or not to leave it enabled (for now, since the next version will likely remove that choice altogether).

But maybe this is best resolved by just waiting until we finish up the VPAT. If we discover an issue within the Job Composer we can list it here as a reason to disable it, if not then we don't need a note.

johrstrom:
Given the conversation today I wonder what you're thinking is on this now?

Bubballoo3:
Yeah I think with how we addressed the consistent navigation rule in the VPAT we can do without a note.

PR #1376, opened 2026-05-13
New tab

### LINKED ISSUE #1325: add docs for new_tab functionality with regard to WCAG compliance
WCAG guidelines say you shouldn't open links in new tabs, so this documentation should reflect that/also say the same.

This here from the customization page as an example:

```yaml
nav_bar:
  # Single top level link
  - title: "Title"
    icon: "fa://desktop" # optional, defaults to fas://cog
    profile: "team1"
    new_tab: true # optional, defaults to false
```

Note this also applies to the `manifest.yml` documentation page, wherever that is.  Related to https://github.com/OSC/ondemand/issues/5115

johrstrom:
https://osc.github.io/ood-documentation-test/new-tab/

Fixes #1325 by documenting the behavior change and updating the accessibility recommendations for the same.

### DIFF
--- a/source/accessibility_settings.inc
+++ b/source/accessibility_settings.inc
@@ -12,6 +12,7 @@ as they may be site specific.
   to complete this criteria.
 * Links should open in the current tab or give warning when they open in a new tab or window.
   This is to adhere to `WCAG Technique G200 "New Tabs"`_ and `WCAG Technique G201 "Advanced warning for new tabs"`_.
+  Note this applies any custom :ref:`menus-for-custom-pages` your center may have created.
 
 .. _WCAG: https://www.w3.org/WAI/standards-guidelines/wcag/
 .. _WCAG criterion 3.2.6 "Consistent Help": https://www.w3.org/TR/WCAG22/#consistent-help
--- a/source/accessibility_settings.inc
+++ b/source/accessibility_settings.inc
@@ -10,8 +10,12 @@ as they may be site specific.
   to complete this criteria.
 * `WCAG criterion 3.2.6 "Consistent Help"`_: Centers need to :ref:`extend the help menu <help_menu_guide>`
   to complete this criteria.
-
+* Links should open in the current tab or give warning when they open in a new tab or window.
+  This is to adhere to `WCAG Technique G200 "New Tabs"`_ and `WCAG Technique G201 "Advanced warning for new tabs"`_.
 
 .. _WCAG: https://www.w3.org/WAI/standards-guidelines/wcag/
 .. _WCAG criterion 3.2.6 "Consistent Help": https://www.w3.org/TR/WCAG22/#consistent-help
 .. _WCAG criterion 2.4.5 "Multiple Ways": http://www.w3.org/TR/WCAG20/#navigation-mechanisms-mult-loc
+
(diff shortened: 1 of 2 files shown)


source/release-notes/v4.2-release-notes.rst
```diff
@@ -143,6 +143,12 @@ and copying the file contents, effectively downloading the file. With this chang
 users from exporting the contents of any files, while still being able to execute certain files in predefined contexts,
 like batch connect apps or :ref:`templated projects <project_templates>`.
 
+Links open in current tab
+.........................
+
+Links will now open in the current tab. This changes the default behavior of previous versions
```
Bubballoo3:
```suggestion
All default links will now open in the current tab. This changes the default behavior of previous versions
```

I think this wording will hopefully call people's attention to the non-default links that they may have to reconfigure

source/release-notes/v4.2-release-notes.rst
```diff
@@ -143,6 +143,12 @@ and copying the file contents, effectively downloading the file. With this chang
...
+that generally opened links in a new tab.
```
Bubballoo3:
```suggestion
that opened some links in a new tab.
```

I think this is slightly more accurate. IIRC it was typically when links would jump from one core app to another like 'Open in files app' or during file editing/viewing

source/accessibility_settings.inc
```diff
@@ -10,8 +10,12 @@ as they may be site specific.
   to complete this criteria.
 * `WCAG criterion 3.2.6 "Consistent Help"`_: Centers need to :ref:`extend the help menu <help_menu_guide>`
   to complete this criteria.
-
+* Links should open in the current tab or give warning when they open in a new tab or window.
```
Bubballoo3, line 13: This should mention/link to the navbar customization at https://osc.github.io/ood-documentation/latest/customizations.html#menus-for-profile-links as well as any place that this is set in an app's config.

Bubballoo3: Specifically I am thinking about 
```
app/views/products/show.html.erb:38:            target: link.new_tab? ? "_blank" : nil
```
which reads the link from `@product.app.links_without_validation`, but not sure where that is configured

johrstrom: > This should mention/link to the navbar customization at https://osc.github.io/ood-documentation/latest/customizations.html#menus-for-profile-links as well as any place that this is set in an app's config.

I noticed that too, but they default to false. Are you suggesting updating those example YAML structures' comments?

Bubballoo3: No, I mostly just want to show people which configs they would have to update to disable all new tab links on their site. Even though they default to false, showing them the relevant files allows them to check and verify that nothing is overriding the default.

So I don't think we need to update those docs (although we could note the possible concerns in a warning), but especially since there are only 1 or 2 locations where a new_tab setting exists, I think it will be easier for us to link than for people to find. Although maybe there is a simpler way like `grep -r 'new_tab' /etc/ood/config/ondemand.d/`. Which I am fine with if we can verify that is the only place it could be set

johrstrom: OK - added a note on that. Let me know what you think.

Bubballoo3 (approved):
LGTM

Issue #1377, opened 2026-05-13
Review Upgrade instructions in release notes


moffatsadeghi:
- Dev to review upgrade instructions to confirm accuracy and clarity
- Confirm all OSes are represented in the release notes 
- Remove any that are no longer supported
- Add any that are now supported

johrstrom:
I think this is complete, as in we're good on these items.

PR #1379, opened 2026-05-13
add new OS steps to RNs

johrstrom:
https://osc.github.io/ood-documentation-test/42-oses/

Work on #1377 to add Debian 13 and Ubuntu 26.04 steps to the release notes.


treydock:
We have historically not added new OS upgrade instructions to release notes because no one has OnDemand 4.1 on those OSes so there is no upgrade process for Debian 13 and Ubuntu 26.04.

johrstrom:
> We have historically not added new OS upgrade instructions to release notes because no one has OnDemand 4.1 on those OSes so there is no upgrade process for Debian 13 and Ubuntu 26.04.

🤦‍♂️ yea that makes sense.

PR #1384, opened 2026-05-14
add note for custom pages and Passenger applications

### LINKED ISSUE #1339: Add release note warning about widgets and other custom content
We should make sure we mention that while the published VPAT covers the dashboard, the job composer, widgets, and other custom apps are not covered and individual centers should review these optional and custom features for accessibility compliance

johrstrom:
https://osc.github.io/ood-documentation-test/accessible-apps/

Fixes #1339, unless we want this in the release notes as well? I just added it to the accessibility page.


Bubballoo3 (approved):
LGTM

PR #1385, opened 2026-05-14
update accessibility.rst and release notes with vpat link

### LINKED ISSUE #1327: ACR / VPAT for 4.2
Produce a ACR as part of Open OnDemand version 4.2.  

Website will need to be updated here https://www.openondemand.org/accessibility with the most recent version.  Maybe we keep version 4.1 and add the version 4.2 so both are available to view and grab?

johrstrom:
https://osc.github.io/ood-documentation-test/add-vpat/

Add a VPAT link the the release notes and accessibility.rst.

Fixes #1327

### DIFF
--- a/source/accessibility.rst
+++ b/source/accessibility.rst
@@ -11,7 +11,17 @@ Open OnDemand has undergone several reviews and updates to meet
 take to achieve complete compliance.  These steps are documented here along
 with areas or certain features that are not complaint.
 
+Accessibility Conformance Report 
+--------------------------------
+
+Developers have uploaded the `Open OnDemand 4.2 VPAT`_ online which
+you may retrieve and review. This document is from the voluntary product assessment
+template a registered service mark of the Information Technology Industry Council (ITI).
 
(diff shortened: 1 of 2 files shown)


source/accessibility.rst
```diff
@@ -11,7 +11,17 @@ Open OnDemand has undergone several reviews and updates to meet
 take to achieve complete compliance.  These steps are documented here along
 with areas or certain features that are not complaint.
 
+Accessibility Conformance Report 
+--------------------------------
+
+Developers have uploaded the `Open OnDemand 4.2 VPAT`_ online which
```
Bubballoo3: It seems like this page should be more generic than it is, mentioning that a VPAT is and will be published on all releases >4.0, and directing to https://www.openondemand.org/accessibility, so that we don't have to update it again in the future

johrstrom: OK

Bubballoo3 (approved):
Made a small edit to wording but now this looks good!

Issue #1401, opened 2026-07-17
Migrate source files from .rst to .myst


Bubballoo3:
This is an epic ticket for the broad migration from reStructuredText to myST, which will allow markdown syntax to work within the docs.

PR #1406, opened 2026-07-28
initial conversion of accessibility page from rst to myst

### LINKED ISSUE #26: Add MyST compatibility to container

Bubballoo3:
Contributes to #1401 and fixes https://github.com/OSC/ood-documentation-build/issues/26. This PR will only work if you open and build the updated container in https://github.com/OSC/ood-documentation-build/pull/27. With that container, this change adds myst_parser to the sphinx extensions in `conf.py` and demonstrates the conversion from rst to myst syntax. I started with the accessibility page as it was quite short while still demonstrating how titles, links, and more complex structures like inclusions get handled in myst. 

Some notes
- With the myst extension enabled, files will parse as rst or myst depending on the extension. For this reason, I had to rename the README.md under architecture, as it was not intended to be a documentation page.
- note that internal links via `ref` still function, the only difference is the demarkation (braces instead of colons)
- One slight change occurs with link definitions. With rst we had the `'wcag'_` definition in accessibility_settings.inc, which could then be used by any parent file that included it. With myst, it is reversed so that the included file can use links defined in parent files, but not vice versa.


source/accessibility.md
johrstrom, line 1: You should do a `git mv` for these files to preserve the git history.

johrstrom:
Now that I can see the changes here - I wonder what's the real difference here? Links are still strange syntax, directives are still a strange syntax. Is all this work really just so heading are slightly easier? What else are we gaining by all of this?

source/accessibility.md
Bubballoo3: This should be resolved now. It still shows file deletion/added in the PR, but if you git blame the file in the branch you can see the full history there.

Bubballoo3:
>  Is all this work really just so heading are slightly easier? What else are we gaining by all of this?

I think the biggest thing we gain is that people are much more likely to have prior experience with markdown, and so many of the smaller formatting/content changes can be made by community members without them needing to explore the all the myst syntax extensions. So while headers, italics, bolds, and code blocks might feel like a minor improvement, that easily covers most of the changes people will want to make, and they will be able to do so using existing intuition instead of running into difficulty with doing the same simple formatting through rst. Even core team members (I for one) have had difficulty with formatting in rst, so this is a substantial win for everyone.

Ultimately I think the myst solution is a great direction because it seems to understand the different audiences and contributors who the documentation source is for. A novice contributor wants something familiar and approachable, while the core developers are willing to learn something complex to avoid repetition and keep things maintainable. As seen in this PR, Myst will allow us to keep all of our existing tools and structures that we like to use, while making the actual content formatting more approachable to a general audience. 

Other proposals, like MkDocs, would have required moving to pure markdown syntax and sacrificing links and directives. So this is a compromise that (IMO) has the potential to address the biggest needs of both groups. Definitely open to any feedback or alternative suggestions on how we could satisfy those two groups better!

Bubballoo3:
I had to add 'html' to the spelling list for the internal link `<change_welcome_html>` though not exactly sure why, since the only change to the link was `:ref:` to `{ref}`. Is there somewhere we can configure what gets ignored by spellchecking?

johrstrom:
> Is there somewhere we can configure what gets ignored by spellchecking?

From the docs https://sphinxcontrib-spelling.readthedocs.io/en/latest/customize.html there appears to be patterns we can ignore, but I think it's fine to just add `html` to the list.

`spelling_exclude_patterns=['ignored_*']`

johrstrom (approved):
LGTM

PR #1412, opened 2026-08-21
correct this default

johrstrom:
This default is true as the section indicates.

https://osc.github.io/ood-documentation-test/johrstrom-patch-1/

This default is `true` as the section below it indicates.


moffatsadeghi (approved):
Correcting the default - docs say the default is false but its actually true.  Blue line in the docs is being changed

PR #1414, opened 2026-08-27
update build from source docs

### LINKED ISSUE #1410: Update Building from Source
https://github.com/OSC/ondemand/blob/master/Dockerfile.example this call out needs removed as a link and sentence. 

Update third link on opening an issue to our new issue template.

johrstrom:
https://osc.github.io/ood-documentation-test/update-src-build/

Fixes #1410  by updating building from source documentation. For whatever reason, this was left in two different spots, so this removes one of the entries.

### DIFF
--- a/source/installation/install-software.rst
+++ b/source/installation/install-software.rst
@@ -212,13 +212,12 @@ Building From Source
 Building from source is left as an exercise to the reader. 
      
 It's not particularly difficult to build the code, but installing it with all the various files is. Should you be interested, 
-review the ``Dockerfile`` and packaging specs for what would be involved.
+review the packaging specs and rules for what would be involved.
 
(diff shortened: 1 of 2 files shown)


source/installation/install-software.rst
moffatsadeghi, line 209: Thanks for updating this section! Since you are already here, I've found this section to be a bit subjective. Can be make it be a little more neutral? Here are some recommendations for how we can update the whole section:

`Detailed instructions for building from source are not provided here.`

`If you're interested in building from source, review the packaging specs and rules for more information about what is involved.`

johrstrom: Done.

moffatsadeghi (approved):

PR #1416, opened 2026-09-03
add docs for html_options

### LINKED ISSUE #128: Document html_options and other enhancmenets to interactive app plugins
`html_options`, the fact that you can add data attributes to select options, etc.

It might be best to have examples of common use cases, mixing this and the javascript file to include.

johrstrom:
https://osc.github.io/ood-documentation-test/html-options/

Fixes #128 by adding this section to the `form.html` page.

image

### DIFF
--- a/source/how-tos/app-development/interactive/form.rst
+++ b/source/how-tos/app-development/interactive/form.rst
@@ -623,6 +623,29 @@ are:
 
           step: 1
 
+.. describe:: html_options (Hash, null)
+
+    Supply HTML options to the element.
+
+    Default
+      No options given.
+
+      .. code-block:: yaml
+
+        html_options: null
+
+    Example
+      Add to the HTML classes of the element to change the color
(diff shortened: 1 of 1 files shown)


Bubballoo3 (approved):
LGTM. In testing I noticed that the quotation marks below are not strictly necessary, but I don't think there is a consistent pattern to force us one way or another. I would be fine either way, so I'll leave it up to you

johrstrom:
> LGTM. In testing I noticed that the quotation marks below are not strictly necessary, but I don't think there is a consistent pattern to force us one way or another. I would be fine either way, so I'll leave it up to you

I feel like I consistently use them, but the project/documentation may not. They are not strictly necessary in YAML, but it's just one habitual patterns I've picked up to avoid having to use it specifically for strings with YAML characters like `:` or similar.

(as an aside I was surprised to see the i18n tasks/lint remove them)

PR #1418, opened 2026-09-08
document submit_host_override for lha

### LINKED ISSUE #251: Document LHAs submit_host_override script.native parameter
A feature was developed out of the discourse issue below. This ticket is to document that feature now that it is a part of the core library.

https://discourse.osc.edu/t/ood-server-offload/625/11

Specifically it's the ability to override the submit host in the job (overridden from what's in the adapter).

```ruby
  def submit_host(script = nil)
    if script && script.native && script.native['submit_host_override']
      script.native['submit_host_override']
    else
      @submit_host
    end
  end
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737368881204) by [Unito](https://www.unito.io)

johrstrom:
https://osc.github.io/ood-documentation-test/lha-host-override/

Fixes #251 by documenting `native.submit_host_override` for the linux host adapter.

Visually this provides this HTML:
image


moffatsadeghi (commented):
Reading through the Discourse post, there is something that gets brought up about the ssh_hosts:

"The overridden host must be in the ssh_hosts list." 

should we mention the relationship with ssh_hosts here too? It sounds like any hosts this can resolve to need to be included in ssh_hosts so OOD can find/manage the job afterward.

Based on the discourse, this seemed like useful context to include with the example if that's still the case... because this is from like 2019

johrstrom:
> should we mention the relationship with ssh_hosts here too? 

Can't hurt. let me look it up. I believe it's still relevant.

johrstrom:
Looking at this page a bit more, the order seems to be off. Generally all adapter have configuration first with extra information below that. This has some extra information above 

Adding a table of contents demonstrates this. cluster configuration being the yellow highlight and this addition being the blue.
image

I would like to also fix that in this PR, do you think that's OK?

moffatsadeghi:
Sure, why not!

johrstrom:
> "The overridden host must be in the ssh_hosts list."

Added a warning section with a link to the relevant part of the page on debugging.

image

> Sure, why not!

Added a ToC and rearranged the page so that the configuration is first.

PR #1420, opened 2026-09-10
document advanced ondemand.d global configs

### LINKED ISSUE #250: Document trick to sharing code between interactive apps
See https://discourse.osc.edu/t/where-to-put-boilerplate-code-used-in-templates/826 and https://listsprd.osu.edu/pipermail/ood-users/2018-November/000297.html

johrstrom:
https://osc.github.io/ood-documentation-test/global-items-w-initializer/

Fixes #250 

That ticket predates global configs, so it doesn't mention it. In any case, using `global_bc_form_items` is the preferred method to do this so that's why this is under that heading.

The addition produces this HTML (zoomed out quite a bit to get it all in one image)

image


moffatsadeghi (changes requested):
Good find that this is missing from our docs, first of all.  Second, I think something that could be even more helpful is to give another example besides the two examples of accounts.  So say, queues, partitions, etc.  
So folks can see the reusable pattern.  Probably a good standard to do for our examples since OOD can be difficult to understand and not everyone is familiar with Ruby on Rails.

johrstrom:
> I think something that could be even more helpful is to give another example besides the two examples of accounts. So say, queues, partitions, etc.

I can add more text about what else could be possible, but I'm not sure if more examples provide value.  Another example would only replace the name `accounts` with `queues`, otherwise they'd be exactly the same. They wouldn't actually _demonstrate_ anything new.

moffatsadeghi:
> Another example would only replace the name `accounts` with `queues`, otherwise they'd be exactly the same.

I see what yo are saying but my concern is still valid.  The problem isn't that people need to see the code twice with different names.  Its that they often  need to understand the scope of how they can apply it.  

If you want, we can keep the examples as is, but add a line or two that the same pattern works for any dynamic global form item.  Whether it be queues, partitions, project codes, etc.  Swap the script and item name.

johrstrom:
> I see what yo are saying but my concern is still valid. The problem isn't that people need to see the code twice with different names. Its that they often need to understand the scope of how they can apply it.

I agree, I've updated the text to expand on what else could be done.

moffatsadeghi (approved):
Thanks, Jeff.  Appreciate you finding a middle ground here!

moffatsadeghi:
I'll let you squash, merge, and delete branch when you want.

